PR #96, opened 2021-02-19
Japanese translation from tokchin

tokchin:
jp translation done.


yangwao:
awesome, don't forget to post your KSM address!

yangwao:
going to check, if its ok to publish live! but so far seems yes!

PR #102, opened 2021-02-19
Update urls to use kodadot repository

JKrupinski:
This pull request updates README file to use urls from kodadot repository


yangwao (approved):
nice one

Issue #83, opened 2021-02-18
Show price of $KSM somewhere in the UI


yangwao:
Would be nice to mind users price for to coin what they are looking for, obv user story is they are checking price on some sort of coingecko or so. We can have it in our UI and avoid user context switching.

yangwao:
Yup, done!

PR #175, opened 2021-03-10
Extend Money Component to show fiat value

JKrupinski:
This is related to #83

Where do you want to show kusama price?
Can I use coingecko api to get it? (Rate Limit: 100 requests/minute)

For now it looks like this (bottom right corner)
alt test


vikiival:
Hey @JKrupinski, 

I would extend the `Money` component with a prop `showFiatValue` which would take fiat compatible argument for coingecko (it think eur / usd) .

Fiat value is important in the NFT detail.
Screenshot 2021-03-10 at 22 22 54

JKrupinski:
@vikiival Do you mean extend '@/components/shared/format/Money.vue'; component, add showFiatValue as a prop and import <Money /> in my KusamaPrice component? or am I wrong

yangwao:
@JKrupinski yup, I guess so. 

I mean showing price without context doesn't make any sense, so having it next to the value of NFT would give most of contextual sense to me :)

JKrupinski:
@vikiival please review my code
it will look like this:

alt gallery item

dashboard/src/components/shared/format/Money.vue
```diff
@@ -25,6 +35,33 @@ export default class Money extends Vue {
     return this.chainProperties.tokenSymbol
   }
 
-}
+  public created() {
```
vikiival, line 38: use `mounted` instead of `created`.

dashboard/src/components/shared/format/Money.vue
```diff
@@ -25,6 +35,33 @@ export default class Money extends Vue {
...
+    if (this.showFiatValue) {
+      this.getFiatValue()
+    }
+  }
+
+  private async getFiatValue() {
+    try {
+      const { data } = await axios.get(`${this.apiUrl}/simple/price`, {
```
vikiival, line 46: I would move this to separate file `src/utils/coingecko.ts` or `src/coingecko.ts` or under `src/fetch.ts`

dashboard/src/components/shared/format/Money.vue
```diff
@@ -25,6 +35,33 @@ export default class Money extends Vue {
...
 
+      this.fiatValue = data[this.coinId][this.showFiatValue] * Number(this.value)
+    } catch (error) {
+      console.log(error)
+    }
+  }
+}
 </script>
+
+<style lang="scss">
```
vikiival, line 61: Wow nice !

vikiival (changes requested):
Please fix those minor things. 
On the other hand you made awesome job!

dashboard/src/components/shared/format/Money.vue
JKrupinski: changed but I don't know why do you want to use `mounted` here
with `created` hook we can get data faster

dashboard/src/components/shared/format/Money.vue
JKrupinski: moved to `src/coingecko.ts` please check if it is that you want, I've added just new axios instance with default url to coingecko

JKrupinski: If I misunderstood it and you wanted to move whole request to separate file just let me know

vikiival: Yeah I just wanted to move it so separate file

dashboard/src/components/shared/format/Money.vue
JKrupinski: simple solution :smile:

dashboard/src/components/shared/format/Money.vue
vikiival: I always used mounted for fetching. However, it looks like it doesn't matter.
https://stackoverflow.com/a/57089636

vikiival (approved):

dashboard/src/components/shared/format/Money.vue
JKrupinski: I always use `created` for fetching because it's executed before mounted and I don't need DOM elements here

Issue #161, opened 2021-03-07
Share to instagram -> Share to Facebook


yangwao:
Maybe instagram has some URI, would be nice to add button to share is as insta story?

BrightReach:
I'd like to give that a go. Whereabouts would you like implement the feature? How do you want it to work?

yangwao:
Well, we want to be generally complaint with sharing to Instagram, thus I guess now it works as posting it as story or post?

JKrupinski:
Looks like it possible only for Android/iOS
https://developers.facebook.com/docs/instagram/sharing-to-feed
https://developers.facebook.com/docs/instagram/sharing-to-stories

I didn't find anything which can help share data to instagram from a web page

yangwao:
Oh probably we can pass on this if it's hard to no-near to make it :)

Or probably choose other popular social network, maybe facebook would be good?

yangwao:
it's there 
![image](https://user-images.githubusercontent.com/5887929/115385479-7a316f00-a1d8-11eb-83ea-28b12d5afa8a.png)

Closing

PR #286, opened 2021-04-10
Add vue-social-sharing plugin

JKrupinski:
Related to #161

My proposition is to add vue-social-sharing package. I've added `ShareNetwork` component which helps share link to a facebook.
Twitter, Telegram, Line can be added too. With this plugin you can share content without leaving kodadot page.

More details here: https://github.com/nicolasbeauvais/vue-social-sharing

You can see new facebook icon in the nft detail page
alt text

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -51,6 +51,7 @@
     "vue-meta": "^2.4.0",
     "vue-property-decorator": "^8.4.0",
     "vue-router": "^3.1.5",
+    "vue-social-sharing": "^3.0.7",
     "vuex": "^3.1.2",
     "vuex-persist": "^2.2.0"
   },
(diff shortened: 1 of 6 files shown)


yangwao:
I was thinking.. what if we could make a show in a default state with just a Twitter button and three dots?
Under three dots could expand into rest buttons?
The purpose of this would be to preserve and keep users low on brain load. 
Towards cleaner and highlighting only buttons which are more used for now :)

Dots could be, probably horizontal fit better?
https://fontawesome.com/icons/ellipsis-v?style=solid
https://fontawesome.com/icons/ellipsis-h?style=solid

yangwao:
Oh missed the first part.

Sure, I'm up to adding vue-social-sharing.

vikiival:
Yeah, It would be awesome to have a plugin for that

JKrupinski:
Good idea @yangwao I think we can hide them all under `share-alt` icon

yangwao:
I mean, I would add even embeding button and leave just copy link and share, to cut on ones brain load :)

Pic from Twitter

IMG_20210415_094215

PR #288, opened 2021-04-12
great filter

vikiival:
We can search more. 
Currently additions.

Sort by something.
Filter by types.

It's not perfect, however a lot of code can be reused

Screenshot 2021-04-12 at 15 34 39


yangwao:
Are we almost past for it? 
https://mason.gmu.edu/~rhanson/greatfilter.html

PR #296, opened 2021-04-17
Change loader styles

JKrupinski:
My proposition is to change loader styles to center it and fix it on mobile view


yangwao:
works like a charm!

PR #316, opened 2021-04-26
gallery optimization

JKrupinski:
To avoid too many requests I had to change pausing gifs to use nft type property (which is mime type)

It works but looks like this attribute is added only for the newest NFTs

Second thing is that Money component makes too much requests to coingecko on gallery page, I've changed it to use component lazy loading and now it sends requests to coingecko only when NFT is visible

But I have another idea, maybe we can change Money component to get KSM price only once and save it in vuex? 
So there will be only one request to coingecko on gallery page


vikiival:
> @vikiival still not sure about formatBalance decimals, I saw that you changed it to 2

Yeah It's highly possible, I think it didn't work on gallery

Meanwhile it was fixed by #acc45dff2299b2580dcb5d6ee64f14770f524c26

JKrupinski:
@yangwao please have a test and merge if everything looks good

vikiival:
> Not sure why some of them don't have it, any ideas?

They were lost in the echo.
Honestly I don't know, I can remigrate those without type.

PR #354, opened 2021-05-08
Generate social cards on the fly with NFT name, price and KodaDot branding

AndreyGrin:
Here image preview should already work. 
But to make it really nice, I will require some help, probably from @vikiival 

1. How can I get `nft.price` in the same view as <money> component returns but I need only value as string.
2. Is it possible to get `width` and `height` attributes for images? This will allow me to build previews based on image sizes and not with fixed dimensions.

### DIFF
--- a/dashboard/src/components/rmrk/Gallery/GalleryItem.vue
+++ b/dashboard/src/components/rmrk/Gallery/GalleryItem.vue
@@ -85,6 +85,7 @@ import 'markdown-it-vue/dist/markdown-it-vue-light.css'
 import { NFTWithMeta, NFT } from '../service/scheme';
 import { sanitizeIpfsUrl } from '../utils';
 import { emptyObject } from '@/utils/empty';
+import { formatBalance } from '@polkadot/util';
 
 import AvailableActions from './AvailableActions.vue';
 import { notificationTypes, showNotification } from '@/utils/notification';
@@ -104,6 +105,7 @@ type NFTType =  NFTWithMeta;
 
 @Component<GalleryItem>({
   metaInfo() {
+    const image = `https://og-image-green-seven.vercel.app/${encodeURIComponent(this.nft.name as string)}.png?price=${this.nft.price ? Vue.filter('formatBalance')(this.nft.price, 12, 'KSM') : ''}&image=${(this.nft.image as string)}`;
     return {
       meta: [
         {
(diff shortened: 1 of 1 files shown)


AndreyGrin:
I think build fails for same reason as I had in local. `vue-social-sharing` package is missing in package.json Should I push updates `package.json` as well?

vikiival:
> Should I push updates package.json as well?
Yes please.

> How can I get nft.price in the same view as component returns but I need only value as string.

I'm not sure that I understand the question.

> Is it possible to get width and height attributes for images? 

[Reading Image Sizes and Dimensions with Vue.js](https://www.raymondcamden.com/2019/06/13/reading-image-sizes-and-dimensions-with-vuejs)

[Vue: get image width + height](https://stackoverflow.com/questions/59253669/vue-get-image-width-height-returns-zero-on-component-load)

yangwao:
Just testing out
Telegram works perfect! 
image
Twitter is probably using other format? 
image
Discord plays a well! 
image

AndreyGrin:
Tested in twitter. It doesn't work in their validator but works fine in twitter posts.
Screenshot 2021-05-10 at 11 18 53
@yangwao Do you want to change anything or we are good to go? I would say it looks good with wide format even if it cropped, but this is really stands out and takes attention for my opinion.

vikiival:
Wow nice ! 🚀

yangwao:
> Tested in twitter. It doesn't work in their validator but works fine in twitter posts.
> Screenshot 2021-05-10 at 11 18 53
> @yangwao Do you want to change anything or we are good to go? I would say it looks good with wide format even if it cropped, but this is really stands out and takes attention for my opinion.

Wow, I could not add more positive emotions to this! I like it! As first iteration on this big chunk is lot of stuff :)
Also never realized that twitter has X on the left in preview mode. 

But hey, what about the text when it's posted, it's cropped?

Just noted that Medium is doing some previews as well 
image

AndreyGrin:
Changed aspect ratio to be 2:1 as in twitter doc, but now it cuts on sides.
Screenshot 2021-05-10 at 16 11 40 How could I test medium? I tried to use link in post and comment but it doesn't build preview

yangwao:
> Changed aspect ratio to be 2:1 as in twitter doc, but now it cuts on sides.
> Screenshot 2021-05-10 at 16 11 40 

I mean perfect, let's put logo slightly far from edge and it would be perfect to fly! :)
Yet I see you need to handle undefined there :)

> How could I test medium? I tried to use link in post and comment but it doesn't build preview

Log into medium
open https://medium.com/new-story 
paste link
hit enter
take a while <- this seems really random

and boom
but sometimes, takes longer than.. while.

image

Anyway, seems medium doesn't have tool for themselves, but have links to other debugers 
https://help.medium.com/hc/en-us/articles/215769058-Update-social-media-preview-cards

Preview from linkedin 
image

AndreyGrin:
Screenshot 2021-05-10 at 20 20 28 Yay, twitter finally good. Pushed price and logo few pixel more from sides to avoid cutting at other resources. 
For a note: while testing, use different images, since Vercel caches image forever, so it won't regenerate it with new size/styles. Fixed undefined price. The only thing what's left is a correct price display in KSM/USD format. Who can help with that?

yangwao:
@vikiival can guide you through, there is a format balance component and we would like to show just KSM as USD may change since then when was it was first cached.

What's are options for vercel pricing or it's free tier I guess for now?
I mean, you can then probably transfer the repository to kodadot and we can deploy it? )
Love to have your long-term contributions :)

vikiival:
@AndreyGrin we use this [formatBalance function](https://github.com/kodadot/nft-gallery/blob/acc45dff2299b2580dcb5d6ee64f14770f524c26/dashboard/src/utils/formatBalance.ts)

so if you call `{{ value | formatBalance }}` it should be allright

AndreyGrin:
Got error: `You don’t have the permission to create public repositories on kodadot` so probably I need transfer it to user. Should I transfer to you @yangwao ?

yangwao:
Yes, @AndreyGrin  you can transfer it to me, I'll make deploy on vercel and we can let it merge into main and probably push to the production :)

yangwao:
okay, we hope that vercel won't limit us but eventually we can afford migrate under organisation later, sometime :)

Issue #243, opened 2021-03-28
Seo upgrade 


yangwao:
# Tasks to do 

- [ ] generate sitemap.xml on build from list to include application pages
  - [ ] rmrk/rare
  - [ ] rmrk/gallery
  - [ ] rmrk/mint
  - [ ] rmrk/create
  - [ ] rmrk/credit
  - [ ] FAQ
  - [ ] Settings
  - [x] Carbonless - define og:description and og:title
  - [ ] Sustainability - define og:description and og:title
- [ ] tune seo description for NFTs to reflect 
  - [x] put nft token name in title
  - [x] description of NFT to the description of page
  - [x] fill og:author with creator address
- [ ] deliver screenshots 
  - [ ] preview how it looks like for indexed sitemap
  - [ ] preview of NFT detail 
- [ ] post link tools you used to test your seo
https://www.google.com/search?q=site%3Anft.kodadot.xyz

https://user-images.githubusercontent.com/5887929/112744027-ab819b00-8f9c-11eb-9bc7-566e9638c7ea.mp4

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.3775 KSM (150.63 USD @ $399.01/KSM)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/kodadot/nft-gallery/243/100025705).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/kodadot/nft-gallery/243/100025705).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.com/invite/83BK5z55yB/'>Gitcoin's Discord</a>
* $2,485,346.74 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 265 years, 6 months from now.
Please review their action plans below:


**1) [florian911](https://gitcoin.co/florian911) has started work.**

Yesresponse will be reflected in a comment on the Github issue.)

 
Yes

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/kodadot/nft-gallery/243/100025705).

[...]

PR #409, opened 2021-05-28
Sitemap and meta Title and descriptions

AndreyGrin:


AndreyGrin:
I see Netlify don't want to build now. My idea is to generate sitemap every time we make a build. If it's not necessary, we can manually generate it using `yarn run sitemap` when we want to update. Sitemap have mostly static pages so it will be ok to have it like that. Let me know what do you think.

yangwao:
> I see Netlify don't want to build now. My idea is to generate sitemap every time we make a build. If it's not necessary, we can manually generate it using `yarn run sitemap` when we want to update. Sitemap have mostly static pages so it will be ok to have it like that. Let me know what do you think.

I mean if we can save some build minutes and lay off stuff which doesn't needed everytime, I'm okay with it to be static and upon manually run :)

yangwao:
Closes #243

yangwao:
Interesting on https://www.linkedin.com/post-inspector/inspect/https:%2F%2Fdeploy-preview-409--nftkodadot.netlify.app%2Frmrk%2Fdetail%2F6788636-3CD8C53D036D48B952-CS-CLOUDSCALPING-0000000000000001

image

however I see it here 
image

Issue #356, opened 2021-05-09
missing Unit in simple mint 


yangwao:
**Describe the bug**
Missing unit in simple mint

**To Reproduce**
Go to https://nft.kodadot.xyz/rmrk/mint

**Expected behavior**
Show KSM

**Screenshots**
![image](https://user-images.githubusercontent.com/5887929/117568396-456e5480-b0c0-11eb-9765-cf94db95a663.png)

vutran6853:
It seen to by default the option is blank. In `BalanceInput` Component have `unit` state . One of option is `{ name: '---', value: 1 },` in array. Might cause the bug

yangwao:
@vutran6853 let's make PR, can send you $ :)

chris-saburo:
@yangwao @vikiival @roiLeo  let me figure it out this current issue.

yangwao:
Meanwhile fixed

PR #411, opened 2021-05-28
Should fix missing unit in simple mint

vutran6853:
Should fix missing unit in simple mint Issue #356


yangwao:
Perfect, can you @vutran6853  post your KSM address? :)

yangwao (changes requested):
Oh, you probably meant KSM? 

I mean not sure if this is good way to do it as this is will be used for other than KSM networks

vutran6853:
Sorry for misspell. Made update.

So there another place this will be reuse at?

vikiival:
It's not a good fix. 

@vutran6853 we are trying to be chain agnostic as we would like to reuse this code for metaprime later on .

vikiival (changes requested):
You can still fetch the unit from Vuex

vutran6853:
I don't see anything in vuex for this unit is being use at. Should we have 2 array/lists? One for unit for simple mint and other for metaprime?

vikiival:
Closing as it can be obsolete

PR #421, opened 2021-06-01
Fine grinder for NFT beans

vikiival:

PR #416, opened 2021-05-29
Sustainibility page spanish translation

pandasamanvaya:
This pull request adds the es translation of the sustainability page as mentioned in issue #335.

PR #423, opened 2021-06-04
8k3 changelog

yangwao:
- [x] #421 new subquery backend - huge improvement in load times and speed! thanks to @vikiival 
- [x] #422 before landing redesign would be nice to keep it slim:)
- [x] #416 thanks to @pandasamanvaya


yangwao:
Let it roll :)

PR #434, opened 2021-06-07
Dark mode GalleryItem button

### LINKED ISSUE #433: Buttons on NFT detail should be black as well
I guess with the recent way of going black, let's check where we left white background :)

image

pandasamanvaya:
Fixes #433


yangwao:
LGTM 🚀

PR #441, opened 2021-06-09
Spanish translation of Carbonless page

pandasamanvaya:
Created the Spanish translation of carbonless NFT page using the translation provided by @CdrsAlejandro in #438


yangwao:
Awesome!

Issue #443, opened 2021-06-09
Return to original view-port after pressing esc 


yangwao:
When I have "fullscreen" on NFT detail would be nice to have some shortcut to press ESC (or any shortcut) to revert it back.

Similar like you are exiting full-screen video :)

davidbonnaud:
Hi I'd like to pick up this issue, to be clear do you mean when you click on the fullscreen-view button on the GalleryItem component?

yangwao:
yes, works, closing #444

PR #444, opened 2021-06-10
Added return to original view-port after pressing esc

davidbonnaud:
Added the ability to minimize a fullscreen NFT image by hitting ```esc``` as requested by @yangwao in #443.


yangwao:
works for me, merging!

Issue #449, opened 2021-06-11
Change white shadow on cards to pink glow


yangwao:
Right now when we have black, would be nice to have a pink neon glow behind cards and where else is white.
Seems easy stuff, maybe @pandasamanvaya wants to do it? :)
![image](https://user-images.githubusercontent.com/5887929/121682569-89d48200-cabc-11eb-9606-bff4289daf9d.png)

PR #453, opened 2021-06-13
Added Pink glow to elements

pandasamanvaya:
Changed the white shadow to pink glow for cards and search bar in the gallery and for dropdowns in navbar as mentioned in #449.


yangwao:
Closes #449

PR #489, opened 2021-06-22
fix(profile): show first nft image as metadata

roiLeo:
#465 related
Capture d’écran 2021-06-22 à 9 02 40 AM

PR #493, opened 2021-06-22
added basic og for collection till we fix with proper one, added cont…

yangwao:
…ainer to collection


yangwao:
basically, its working, collection 
image
user profile 
image

we will finish it in #489

PR #525, opened 2021-07-01
fix(gallery): show last nft card

roiLeo:
Last data wasn't loading 

### PR type
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

#### Before
Capture d’écran 2021-07-01 à 1 59 54 PM

#### After
Capture d’écran 2021-07-01 à 1 56 42 PM

### DIFF
--- a/src/components/rmrk/Gallery/Gallery.vue
+++ b/src/components/rmrk/Gallery/Gallery.vue
@@ -163,7 +163,7 @@ export default class Gallery extends Vue {
     type: '',
     sortBy: { blockNumber: -1 }
   };
-  private first = 20;
+  private first = 21;
   private placeholder = require('@/assets/kodadot_logo_v1_transparent_400px.png');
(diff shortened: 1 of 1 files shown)


vikiival (changes requested):
This is not fix. 

The problem is that a NFT can have editions (eg. same picure 5 times on the page).
and we group them by the image. 
So if you have `first` set  to `10`
and one nft which have 5 edition it means that there would be 6 images on the page.

roiLeo:
> This is not fix.
> 
> The problem is that a NFT can have editions (eg. same picure 5 times on the page).
> and we group them by the image.
> So if you have `first` set to `10`
> and one nft which have 5 edition it means that there would be 6 images on the page.

Oh ok didn't know about that.
Is there any example I can find with multiple editions of same NFT ?

vikiival:
Subquery is currently working on groupby feature

Issue #618, opened 2021-08-23
Fix right width for price chart on mobile and tablet breakpoints


yangwao:
Current issue how it looks in ```main``` branch https://github.com/kodadot/nft-gallery/pull/595#issuecomment-903284703

Seems the issue is that breakpoints don't work every time, when you squeeze it in a responsive way, it's okay but when you reload it looks like that, strange :)

PR #622, opened 2021-08-23
fix(gallery-item): history & price chart

roiLeo:
#618, #620, #621 related

#### Before
Screenshot 2021-08-23 at 17-30-46 Marguerite-Marie

#### After
Screenshot 2021-08-23 at 17-32-25 Marguerite-Marie


yangwao:
Wait, at first thought what's change back but I get the flow okay, might work good one :)

PR #624, opened 2021-08-23
fix(gallery): align pagination right

roiLeo:
This small change align pagination in gallery to the right.

### Screenshot
Screenshot 2021-08-23 at 18-44-42 Low minting fees and carbonless NFTs


yangwao:
great eye ball there!

PR #629, opened 2021-08-24
fix(gallery): PriceChart yAxis

### LINKED ISSUE #623: Remove or lighten horizontal lines in price chart
Right now horizontal lines are a bit overwhelming and I guess if we can I would drop any possible cognitive load or pixels

image

Also, "KSM" could be aligned to the left? 
image

roiLeo:
- closes #623


yangwao:
All good, at some point KSM on every line could be somehow deduplicated, but for now is okayish I guess

image

PR #631, opened 2021-08-24
fix: check extension on login page

gdev1001:
#### Fix
* added check extension(polkadot.js) on the login page


yangwao:
You are missing PR template, won't merge till you add it.

I would put it as indefinite notification if so as in three seconds I barely can notice what's was there

image


I would probably put it in snack bar? 
Also, the wording "extension" is missing there
"Inspiration" should be link to "Install Polkadot.js extension" ?
image

PR #695, opened 2021-09-04
separate out logic for twitter

masonwoodford:
**Thank you for your contribution** to the KodaDot NFT gallery.
Let's do a quick check before the merge.
Your contribution will be an eternal part of the community codebase

### Before submitting this PR, please make sure:
- [ ] Code builds clean without any errors or warnings
- [ ] I've merged recent default branch -- **main** and I've no conflicts
- [ ] I didn't break any original functionality

### Optional
- [ ] I've tested it on mobile and everything works

### PR type
- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

- closes #<issue_number> - <Brief_description_how_you_fixed_it>

### If issue contains bounty label ($,$$,$$$)
- Your KSM address: <Kusama_address>

### Community participation
- [ ] Are you at KodaDot Discord? If no, [let's join us - https://discord.gg/35hzy2dXXh](https://discord.gg/35hzy2dXXh)

### Screenshot
- [ ] Whenever your fix will change **something** on UI, a screenshot is more than welcome

### DIFF
--- a/src/views/OnChainProperty.vue
+++ b/src/views/OnChainProperty.vue
@@ -2,6 +2,7 @@
 	<div class="d-flex flex-row">
  		<span v-for="(item, index) in icons" :key="index">
 		<b-tooltip class="p-2"
+		 v-if="icons[index].icon !== 'twitter'"
 		 :label="icons[index].label"
 		  position="is-top"
 		  v-clipboard:copy="icons[index].label"
@@ -10,7 +11,20 @@
 		      :size="icons[index].size"
 		      :pack="icons[index].pack"
 		      :icon="icons[index].icon"
-		  		@click.native="toast('Copied to clipboard')"
+		  		@click.native="toast('Copied to clipboard', icons[index])"
+		    >
+		    </b-icon>
+		</b-tooltip>
+		<b-tooltip class="p-2"
+		 v-else
+		 :label="icons[index].label"
+		  position="is-top"
+		  >
+			<b-icon class="property"
+		      :size="icons[index].size"
+		      :pack="icons[index].pack"
+		      :icon="icons[index].icon"
+		  		@click.native="navigateToTwitter(icons[index].label)"
 		    >
 		    </b-icon>
 		</b-tooltip> 			
@@ -119,7 +133,11 @@ export default class OnChainProperty extends Vue{
 
 	public toast(message: string): void {
 		this.$buefy.toast.open(message);
-  }
+  	}
+
+	public navigateToTwitter(handle: string): void {
+		window.open('http://twitter.com/' + handle.replace(/@/, ''), '_blank');
+	}
 
 	@Watch('legal')
 	async watchLegal(newLegal: string, oldLegal: string){


src/views/OnChainProperty.vue
```diff
@@ -10,7 +11,20 @@
 		      :size="icons[index].size"
 		      :pack="icons[index].pack"
 		      :icon="icons[index].icon"
-		  		@click.native="toast('Copied to clipboard')"
+		  		@click.native="toast('Copied to clipboard', icons[index])"
```
yangwao, line 14: so copy to clipboard and open new window 🤔 I mean get it might be interesting to copy to clipboard if someone wants to share that URL to twitter. but for now having that buttons on right copy to URL of artists profile is enough I guess :)

yangwao (changes requested):
I guess copy to clipboard doesn't work anyways, maybe remove it?

src/views/OnChainProperty.vue
```diff
@@ -119,7 +133,11 @@ export default class OnChainProperty extends Vue{
 
 	public toast(message: string): void {
 		this.$buefy.toast.open(message);
-  }
+  	}
+
+	public navigateToTwitter(handle: string): void {
+		window.open('http://twitter.com/' + handle.replace(/@/, ''), '_blank');
```
roiLeo, line 139: Should be https instead of http

roiLeo (changes requested):

src/views/OnChainProperty.vue
masonwoodford: That second parameter should be removed, sorry. I'm not sure what you mean though, my thought was that if the button is not Twitter than it maintains the previous behavior and if it is Twitter it navigates to the URL. Handled by this logic: 		 v-if="icons[index].icon !== 'twitter'"
So it will not copy to clipboard if it is the Twitter button.

src/views/OnChainProperty.vue
```diff
@@ -13,6 +14,19 @@
 		  		@click.native="toast('Copied to clipboard')"
```
vikiival, line 14: This can be handled by 1 LOC.
`@click= "twitter ? navigateToTwitter : toast"`

vikiival (changes requested):

src/views/OnChainProperty.vue
masonwoodford: fixed

src/views/OnChainProperty.vue
masonwoodford: The only problem with this is that the click method is within the b-tooltip which contains the v-clipboard copy directive. I couldn't find a way to conditionally render that directive. In that case it would both open the new tab and copy the label to the clipboard, which might be OK?

yangwao: ah I guess right now I would just open a new tab and for copy label to the clipboard could be expressed by another button? Because anyway we have there copy link to user profile, so I would omit then copy to clipboard:)

yangwao:
Let me know when we can merge this!

masonwoodford:
> Let me know when we can merge this!

Should be ready to be merged!

Issue #494, opened 2021-06-22
Add Pull Request template


yangwao:
Would help us to answer usual questions
could fix with erroring PR bots

yangwao:
fixed in #508

PR #701, opened 2021-09-08
fix(workflow): gh action lgtm

roiLeo:
IMO pr-chatter should be removed

### PR type
- [x] Bugfix

- fix lgtm action from workflow


yangwao:
There was an issue where I wanted to refine our PR flow 
ah here https://github.com/kodadot/nft-gallery/issues/494

PR #733, opened 2021-09-16
chore: fix lint warning

roiLeo:
- replace double quotes with single quote
-  use 'const' instead of 'let'


yangwao:
ouch, thought we already linting.. is that pre-commit linter hook incoming stadium? 😄

Issue #806, opened 2021-09-27
Not correct preview image NFT


jkdizajn:
the gallery does not correctly display images with an aspect ratio other than 1:1 when opened
Original: https://kodadot.mypinata.cloud/ipfs/QmPgvHLL5rtBtgZA6Rbfjb163wCXtmJtgnbKz7uYHXmbkF
![obrázok](https://user-images.githubusercontent.com/81104034/134897612-8a5449e2-e70d-4837-85a5-950f001d9a94.png)
Galeri preview: https://nft.kodadot.xyz/rmrk/detail/9251128-900D19DC7D3C444E4C-KSMG22-KUSAMAGEN_160-0000000000000001
![obrázok](https://user-images.githubusercontent.com/81104034/134897817-0fad91af-a5a0-4115-8eb1-47ecfcd3a101.png)

yangwao:
Yes, the current stretch is bad, I've put it on p1 as it makes the app unusable and skews artwork.

As I sense that there is **no ratio standard** and anyone can decide what they putting there, how to show it. 
Proposed change would be
- I guess we should crop by default the center middle of art and that's it, preferred square to keep our gallery uniform at some point. We can add it as user settings as everyone can prefer showing it other way around
- When someone triggers Theatre view, it should show non-skewed-stretched-version in the original ratio
- Current view to show the art is possible only through fullscreen mode 
![image](https://user-images.githubusercontent.com/5887929/134899881-f7e07f7c-f9e8-4195-9ded-73551f86c8cf.png)

vikiival:
Fix'd

yangwao:
> Fix'd

Maybe temp fix, but not long-run.

original art is longer than you might see 
https://kodadot.mypinata.cloud/ipfs/QmPgvHLL5rtBtgZA6Rbfjb163wCXtmJtgnbKz7uYHXmbkF

It's okay on item detail now 
![image](https://user-images.githubusercontent.com/5887929/134907349-d671d6bb-8416-435e-8f4d-b4b1990f9721.png)

Not ok on "fullscreen"
![image](https://user-images.githubusercontent.com/5887929/134907438-38012dca-17b3-4a87-9ffc-269064de5488.png)

Not ok on theatre, literally it just zooms now

![image](https://user-images.githubusercontent.com/5887929/134907527-6732513b-25d6-49b3-92db-53c3ae9211e3.png)

Theatre should show original dimensions I guess.

jkdizajn:
[...]

PR #809, opened 2021-09-27
fixed 1 by 1

vikiival:
Closing #806


vikiival:
@jkdizajn said that it's better

yangwao:
Now it's broken even more 😅

vikiival:
> Now it's broken even more
What's wrong with that ?

PR #814, opened 2021-09-27
App should correctly show fullscreen

vikiival:
Thank you for your contribution to the KodaDot NFT gallery, 
we really appreciate your contribution!

### PR type
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

### Before submitting this PR, please make sure:
- [x] Your code builds clean without any erros or warnigns
- [x] You've posted screenshot of demonstrated change in this PR
- [x] Merged recent default branch, **main** and you have no conflicts
- [ ] Didn't break any original functionality 

### Optional
- [x] You've tested it on mobile
- [ ] Are there any edge cases? Name if any

Issue #813, opened 2021-09-27
Full screen and Theatre mode is broken on desktop


vikiival:
**Describe the bug**
Unrelated to #806 as it was broken before

In before #809 :
<img width="1680" alt="Screenshot 2021-09-27 at 14 50 58" src="https://user-images.githubusercontent.com/22471030/134912715-ae34376e-8daa-4dbc-827b-33eb851f127d.png">

<img width="1680" alt="Screenshot 2021-09-27 at 14 51 01" src="https://user-images.githubusercontent.com/22471030/134912691-97e3eb91-5d54-4b6e-9781-023923daab68.png">

vikiival:
After #809 :
<img width="1680" alt="Screenshot 2021-09-27 at 14 59 15" src="https://user-images.githubusercontent.com/22471030/134912992-932833c4-1000-40a5-a39b-0345059fec5f.png">
<img width="1680" alt="Screenshot 2021-09-27 at 14 59 17" src="https://user-images.githubusercontent.com/22471030/134912997-182b5f48-083c-4142-a95d-a26f1f10ddfe.png">

yangwao:
Re #806 - theatre mode is also broken, more pics in this comment https://github.com/kodadot/nft-gallery/issues/806#issuecomment-927824121

vikiival:
Closed in #814

Issue #160, opened 2021-03-07
Limit quantity of NFTs to buy


yangwao:
Sort of anti-whale security to avoid same person can buy whole collection
- [ ] add ability to seller to limit the amount of NFTs that a person can buy

yangwao:
Continue to #3846

PR #815, opened 2021-09-27
fix(GalleryItem): fix image heigh on gallery Item

roiLeo:
Only for testing purposes

### DIFF
--- a/src/components/rmrk/Gallery/GalleryItem.vue
+++ b/src/components/rmrk/Gallery/GalleryItem.vue
@@ -334,7 +334,10 @@ hr.comment-divider {
     margin: 30px auto;
     width: 100%;
 
-    .image {
+    .nft-image img {
+      max-height: 530px;
+      width: auto!important;
+      margin: auto;
       border: 2px solid $primary;
     }
 
--- a/src/components/rmrk/Media/ImageMedia.vue
+++ b/src/components/rmrk/Media/ImageMedia.vue
@@ -2,7 +2,7 @@
   <b-image
     :src="src"
     :alt="mimeType"
-    ratio="1by1"
+    class="nft-image"
   ></b-image>
 </template>
 


src/components/rmrk/Gallery/GalleryItem.vue
```diff
@@ -334,7 +334,10 @@ hr.comment-divider {
     margin: 30px auto;
     width: 100%;
 
-    .image {
+    .nft-image img {
+      max-height: 530px;
```
vikiival, line 338: I think this will be broken on mobile

roiLeo: ### Screenshot
#### Mobile View large image
Screenshot 2021-09-27 at 17-17-55 Decorator Tiles #2 - Dune special
#### Mobile View long image
Screenshot 2021-09-27 at 17-20-29 Kusamagen #160

vikiival:
It would be nice to have a fixed max height so the layout would look the same.
As you can see on the volcano user is able to see descriptions and so on. On the other hand on kusamagen you can see name, collection, and creator (which is correct).

Screenshot 2021-09-30 at 09 43 09

Screenshot 2021-09-30 at 09 43 43

yangwao:
> It would be nice to have a fixed max height so the layout would look the same. As you can see on the volcano user is able to see descriptions and so on. On the other hand on kusamagen you can see name, collection, and creator (which is correct).
> 
> Screenshot 2021-09-30 at 09 43 09
> 
> Screenshot 2021-09-30 at 09 43 43

Make sense to have some fixed height where art is placed. That was original thoughts like foundation is doing it tho. I guess it could be as option or preference for user what they look like to have there :)

yangwao:
Should we take this in or it's already obsolete?

roiLeo:
closed in favor of #814, #813 we will maybe need a fixed height later for GalleryItem

Issue #641, opened 2021-08-25
As Dev I would like to have a standardised code formatting.


vikiival:
**Is your feature request related to a problem? Please describe.**
"I think we should standardized code quality i've seen some files with different style (some with spaces without semicolon others with tabs)"

**Describe the solution you'd like**
Some sort of standardised prettier would be enough for now

**Describe alternatives you've considered**
Have a pre-push hook which lints the code 👁️

PR #820, opened 2021-09-29
chore(linting): migration tslint to eslint

roiLeo:
Big changes over here.

### Usage:
To show problems on all files (.js, .ts, .vue):
```sh 
yarn lint
```
To only show errors:
```sh
yarn lint --quiet
```
Automatic fix:
```sh
yarn lint --fix
```

Right now we've got 1236 problems (0 errors, 1236 warnings)
Next step would be to add this command on our Github Workflow.

### Indentation:
To easily convert **tab** to **space** or **space** to **tab** you will need to edit .eslintrc.js file & .editorconfig
then you can use `yarn lint --fix`

### PR type
- [x] Bugfix
- [x] Refactoring

### What's new? (may be part of changelog)
- This PR is related #641


yangwao:
Seems like this will be adjust, looks good to me!

.editorconfig
```diff
@@ -13,18 +13,18 @@ charset = utf-8
 
 # Indentation override for all JS
 [**.js]
-indent_style = space
+indent_style = tab
```
vikiival, line 16: I would personally use `spaces`

.eslintrc.js
```diff
@@ -1,13 +1,45 @@
...
-    parser: "@typescript-eslint/parser",
-  },
-    plugins: ['@typescript-eslint'],
-    extends: [
-      'eslint:recommended',
-      'plugin:@typescript-eslint/recommended',
-      'plugin:vue/recommended',
-      // 'prettier',
-    ],
+	'env': {
```
vikiival, line 2: format 👁️

vikiival (changes requested):
Spaces ftw

.editorconfig
yangwao: is this evergreen bikeshed discussion on every project 😅

let's go with tabs :D

vikiival: Because she told you that she needs some `space`

I would not mind to use tabs but the code is not properly formatted
Screenshot 2021-09-29 at 13 38 23

roiLeo: Let's agree to use "space: 2", I'm currently fighting to fix "Mixed spaces and tabs" on .vue files.

yangwao: well, tabs could be space 2 tho :)
then there is another discussion about indentation... 2 or 4 😄

yangwao:
For me is good to go!

vikiival (approved):
🚀

Issue #805, opened 2021-09-27
Add addressable paging to the creator profile


yangwao:
Right now when I want to share creator profile at page 5, no one knows where I'm looking at
Current https://nft.kodadot.xyz/rmrk/u/FZpjhYdt6fqRSzW1nR4fRejavcrSo7UXYP4TDhE3mgimgeC 
Final URL could be https://nft.kodadot.xyz/rmrk/u/FZpjhYdt6fqRSzW1nR4fRejavcrSo7UXYP4TDhE3mgimgeC?tab=created&p=5

Issue #587, opened 2021-08-06
Ranking on collections // collectibles


yangwao:
Probably fork spotlight and adjust accordingly, showing in table collections lists, inspiration came from https://opensea.io/rankings, probably create route /rmrk/rankings

Add columns
- [ ] floor price
- [ ] number of items
- [ ] number of owners
- [ ] average price trade for collection
- [ ] total price traded for collection

another inspo https://wgmi.io/
from #586 
> I guess this could be different from the existing spotlight, inspiration is from OS https://opensea.io/rankings
They have Volume, 24h %, 7d%, floor price, number of owners and number of assets in the collection.
I guess if we can figure out it for collections per see, that would be an awesome head start to open doors for collectibles

yangwao:
I guess when we will have stats on collections, now it's @roiLeo time to recycle knowledge and make a dedicated rank component? 👀

roiLeo:
Yup, I can work on that, I'm not a Graphql expert yet but I can manage to make something.

roiLeo:
### Proposal
/rankings page is going to show top 100/50(?)/10(?) collection based on Score

#### (WIP) Columns in order:

- [ ] rank number
- [ ] collection thumbnail
- [ ] collection name linked to user collection
- [ ] floor price
- [ ] volume traded (?)
- [ ] 24h volume
- [ ] 7d volume (?)
- [ ] Number of diiferent owners (?)
- [ ] Number of collected nft
- [ ] Number of nft left (?)
- [ ] Number of nft (total)

*(?) should this be added ?

yangwao:
> Number of diiferent owners 

For sure, important metric of distribution of collection

yangwao:
> volume traded (?)

Yes! :)

> Number of diiferent owners 

For sure, an important metric of distribution of the collection

> 7d volume traded (?)

Yes, you want to see change over time, may be worth put also 2wk basis. Literally NFTs is quite illiquid so people don't trade it that fast is worth nothing on longer scale, inspiration from coingecko 
![image](https://user-images.githubusercontent.com/5887929/134331028-45abde7f-d2e5-40fc-b3f5-fdd3258c72a2.png)

> Number of nft left (?)

Listed for sale but not sold yet? I mean worth considering, in some new collection this could be nice indicator for velocity of distribution if we can set there something along lines of when was first NFT minted?

Issue #586, opened 2021-08-06
Rankings on Collectibles


yangwao:
I guess this could be different from the existing spotlight, inspiration is from OS https://opensea.io/rankings
They have Volume, 24h %, 7d%, floor price, number of owners and number of assets in the collection.

I guess if we can figure out it for collections per see, that would be an awesome head start to open doors for collectibles

yangwao:
continue to #587

PR #829, opened 2021-10-01
Einstellung effect.

yangwao:
linted vuejs, won't be seen by user
addressable tab coming #805 
sort filter is back! #836 
collectibles and rankings #587 #586


yangwao:
Seems build time is around 2m, that's quite good! 

image

PR #819, opened 2021-09-28
#745 Add motion effect on gallery cards in profile

prachi00:
**Thank you for your contribution** to the KodaDot NFT gallery.
Let's do a quick check before the merge.
Your contribution will be an eternal part of the community codebase.

### PR type
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

### Before submitting this PR, please make sure:
- [x] Code builds clean without any errors or warnings
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've didn't break any original functionality
- [x] I've posted screenshot of demonstrated change in this PR

### Optional
- [x] I've tested it on mobile and everything works
- [ ] I found edge cases

### What's new? (may be part of changelog)
- This PR closes #745 - I have created a directive for orientation where I have handled the logic

### If issue had bounty label ($,$$,$$$)
- Fill up your KSM address: 
https://kodadot.xyz/transfer/?target=EzGc4s9PgCPx1YnF3fqzhLzVHpHMTL4LWPScwpDrR8JKgSU

### Community participation
- [x] Are you at KodaDot Discord? If not yet, [let's join us - https://discord.gg/35hzy2dXXh](https://discord.gg/35hzy2dXXh)

### Screenshot
- [x] Whenever my fix has changed **something** on UI, a screenshot for others, is more than welcom
Screenshot 2021-09-28 at 10 07 21 PM
e

PR #916, opened 2021-10-19
feat(galleryItem): image orientation

roiLeo:
#819 related

Orientation is quicker without `el.onLeave()`
Binding Event on `window` instead of `element`

all credit goes to prachi00


roiLeo:
> What's that for?

Testing purpose.

### What changed
Perspective works on full window.
Fixed orientation speed on desktop, see preview.

PR #1408, opened 2021-12-07
🐛 metadata component

roiLeo:
We don't need to watch store state change

### PR type
- [x] Bugfix
- [x] Refactoring


vikiival:
YOLO 🍴

PR #1760, opened 2022-01-10
:bug: Identity popover does not throw 400 for sq

vikiival:
Thank you for your contribution to the KodaDot NFT gallery, 
we really appreciate your contribution!

### PR type
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

### Before submitting this PR, please make sure:
- [x] Your code builds clean without any erros or warnigns
- [ ] You've posted screenshot of demonstrated change in this PR
- [ ] Merged recent default branch, **main** and you have no conflicts
- [x] Didn't break any original functionality 

### Optional
- [ ] You've tested it on mobile
- [ ] Are there any edge cases? Name if any 

### What's new? (may be part of changelog)
- closes #<issue_number> - <Brief_description_how_you've_fixed_it>

### If issue had bounty label ($,$$,$$$)
- Fill up your KSM address: https://kodadot.xyz/transfer/?target=<Kusama_Address>


kkukelka:
i think our PRs fix the same issue,
what do you think about my proposed solution in #1761 ?

vikiival:
@kkukelka my original solution was similar to yours, but

but there are 2 major issues:
1. We use [apollo Smart query](https://v4.apollo.vuejs.org/guide-option/queries.html#queries) there so it's reactive by default we only need to handle initial state when adress is `null` and the query requires any string value
2. When you use `@Watch` the value is firstly reflected in watcher and then after watcher the value is updated


```ts
@Watch('anything')
watchAny(val: string) {
  if (val) {
    this.fetch()
  }
}
```

if you use `this.anything` in `this.fetch` at the time of calling `this.fetch` from the watcher it will have the **old value**

the correct approach is 

```ts
@Watch('anything')
watchAny(val: string, oldVal: string) {
  if (shouldUpdate(val, oldVal)) {
    this.fetch(val)
  }
}
```

kkukelka:
> 1. apollo Smart query

@vikiival  Wow i see your point. didnt know about the smart query capabilities 😮 
in this case i highly prefer your solution here. I will close my PR then

vikiival:
When @yangwao pushes to your branch cbd5d2528daf36c872ff1860fea02d982b362250 and breaks the code

Me:
unknown-10

PR #1787, opened 2022-01-12
📊 vueGtag plugin config

roiLeo:
/giphy rocket


vikiival:
> '~/plugins/vueGtag'

may mode client ?

roiLeo:
> > '~/plugins/vueGtag'
> 
> may mode client ?

Doesn't it work for both?

PR #1822, opened 2022-01-14
🧼 lint service-worker

roiLeo:
- [x] Lint files


yangwao:
I hope we could remove service-worker.js in month or so. probably those who hasn't visited us in month we don't care much

PR #1815, opened 2022-01-14
fix: series insight sorting (client-only)

kkukelka:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).
👇  _ Do a quick check before the merge.

### PR type
- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

https://user-images.githubusercontent.com/17953978/149520423-deb5a5d4-fc31-4eb2-8d7b-b300b05469da.mov

### What's new?
- [x] PR closes #1757
- [x] note that this only fixed client-side sorting; im not aware that backend can handle sorting and only goes by rank right?

### Before submitting Pull Request, please make sure:
- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main-nuxt** and I've no conflicts
- [x] I've tried respect high code quality standards
- [x] I've didn't break any original functionality
- [x] I've posted screenshot of demonstrated change in this PR

### Optional
- [x] I've tested it at </series-insight>
- [ ] I've tested PR on mobile and everything seems works
- [ ] I found edge cases

### Had issue bounty label ?
- [x] Fill up your KSM address: [Payout](https://kodadot.xyz/transfer/?target=EqdyzrzVmeHwMdMwvPeCMnNdbuQDbD3YrjY93xq9Ln3jUGW)

### Community participation
- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

### Screenshot
- [ ] My fix has changed **something** on UI, a screenshot is best to understand changes for others.

Issue #1177, opened 2021-11-16
User based deny list


yangwao:
Let's be honest, we don't like see [spammy feed](https://kodadot.xyz/rmrk/gallery?page=1071&search=&listed=true) and let's add user preferences add filter by
- artist address
- collection address

Make it shareable, so users can share deny lists between them easier. Like generating URL for sharing and load new sharing lists for them

###  Ref
- #942 #1160

yangwao:
Adjusting flight plan 

https://twitter.com/KodaDot/status/1463090371886567433

phillux:
Phil here, thanks for sharing.

A first-draft proposal should research and eventually implement something more like reddit moderation than opt-in user-generated deny lists, if we want to think long-term. 

In order to achieve this, there first needs to be a collectively agreed-upon position of what should be allowed and what should not be allowed on Kodadot. It should be noted that this is a challenge that should be well thought-out, as it is a "continuum problem" that must be shaped over time. For example, we know that images of real-world abuse should always be removed from the platform, and images of cute kittens should always be allowed. So where's the line in between?

Once we are able to establish a code of conduct we think would be appropriate for this community now and in the future, we can then begin to talk about methodologies such as councils, voting thresholds, voting periods, etc...

yangwao:
> Phil here, thanks for sharing.

Welcome!

> A first-draft proposal should research and eventually implement something more like reddit moderation rather than opt-in user-generated deny lists, if we want to think long-term.

What's wrong with that? Happy to learn your perspective. Our long-term plan is we will encourage people to run forks of KodaDot instances in wild. #1171 
 
> In order to achieve this, there first needs to be a collectively agreed-upon position of what should be allowed and what should not be allowed on Kodadot. It should be noted that this is a challenge that should be well thought-out, as it is a "continuum problem" that must be shaped over time. For example, we know that images of real-world abuse should always be removed from the platform, and images of cute kittens should always be allowed. So where's the line in between?

Good Q! Right now I'm looking on that (hopefully media won't pick it up from here) like Aave and save pro, uniswap vs uni swap pro. At some point, we will need to be permissive and limit most of the stuff, however, like uniswap approach, encourage people run their frontends on their own costs of being sued for whatever happened there. Happy learn if there is better tactic, Because we as founder, caring responsibility what is served through "domain-name". 

[...]

PR #1893, opened 2022-01-17
✨ series backend sorting

roiLeo:
Series columns sorting which would need to be fixed later 

### Column not working
- `rank` => ... // TODO backend subsquid

### PR type

- [x] Bugfix

### What's new?

- [x] PR related with #1815 
- [x] PR related with https://github.com/kodadot/rubick/pull/14
- [x] disabled sorting on column that make the gql crash


yangwao:
ooo, related to #1177 ?

roiLeo:
> ooo, related to #1177 ?

yes and no, I tried to integrate the denylist we have but it doesn't seem to accept denylist as a parameter. 
I think we will need to add denyList filter in the new series resolver to make it work.

Issue #1861, opened 2022-01-16
git hooks 


yangwao:
which pre-commit git hooks we should have?

- prettier probably https://prettier.io/docs/en/install.html#git-hooks
- linter?

what's else?

yangwao:
Fixed meanwhile

PR #1914, opened 2022-01-18
⚒️ workflow prettier error

roiLeo:
### PR type

- [x] Bugfix
- changed prettier workflow to run on push

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 
 on:
-  pull_request:
+  push:
     branches: [main]
 
 jobs:


yangwao:
Hmm, best would be having opening code suggestions during PR review? 
Like linter does through review dog?

yangwao:
> > Hmm, best would be having opening code suggestions during PR review? Like linter does through review dog?
> 
> What do you want to achieve with this prettier workflow?

avoid committing non-formatted code, but probably best would be to catch it through git-hooks I guess? 
https://github.com/kodadot/nft-gallery/issues/1861

roiLeo:
closing in favor of #1917

PR #1913, opened 2022-01-18
✨ spotlight backend sorting

roiLeo:
I'm not able to sort on averagePrice column yet.

### PR type

- [x] Bugfix
- [x] Feature

### What's new?

- [x] PR closes #1848 
- [x] backend sorting on Spotlight page

### Before submitting Pull Request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried respect high code quality standards
- [x] I've didn't break any original functionality

PR #1924, opened 2022-01-19
✨ spotlight unsortable column

roiLeo:
### What's new?

- [x] continue #1913  
- [x] disable unsortable column untill we fixed it

### Column not working
- `unique` => message: "syntax error at or near "ASC"" (reserved sql term)
- `uniqueCollectors` => message: "column \"uniquecollectors\" does not exist"
- `count` => message: "column "count" does not exist"

### DIFF
--- a/components/spotlight/SpotlightTable.vue
+++ b/components/spotlight/SpotlightTable.vue
@@ -46,7 +46,7 @@
         <b-skeleton :active="isLoading"> </b-skeleton>
       </b-table-column>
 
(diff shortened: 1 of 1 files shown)


yangwao:
Spotlight columns sorting which would need to be fixed later 

### Column not working
- `unique` => message: "syntax error at or near "ASC"" (reserved sql term)
- `uniqueCollectors` => message: "column \"uniquecollectors\" does not exist"
- `count` => message: "column "count" does not exist"

PR #1967, opened 2022-01-20
Flat Collection fee

vikiival:
:tada: **Introducing the flat fee**:

Now the fee for minting the collection was around $0.07.
Since the last time changed the fee we had: 
- Free Textile
- Free Pinata (only paid for IPFS pins)

Our current stack which we need to pay:
- Cloudflare Images
- Cloudflare Workes
- Cloudflare Pages
- Pinata Pro
- Pinata throughput
- soon or later - Indexers

The basic flat fee is $0.5 per minted collection
For the NFT mint it's also $0.5 + $1 if the NFT is carbonless

- :zap: ability to update images on cloudflare
- :broom: removed useless dependencies from global plugin
- :tada: :construction: new flat fee
- :zap: apply flat fee to the mint components
- :zap: can directly upload image from unique/createCollection
- Some of your tasks use `git add` command. Please remove it from the config since all modifications made by tasks will be automatically added to the git commit index
- :bug: direct upload takes file instead of data
- :loud_sound: 💋 for every support
- :loud_sound: direct upload would throw better error
- :broom: added types for mingo

👇 \_ Do a quick check before the merge.

### PR type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring


components/rmrk/Create/CreateCollection.vue
```diff
@@ -185,7 +174,15 @@ export default class CreateCollection extends mixins(
...
         ? mintInteraction
-        : [asSystemRemark(api, mintInteraction), ...(await this.canSupport())]
+        : [
+            asSystemRemark(api, mintInteraction),
+            ...(await canSupport(this.hasSupport)),
+          ]
+
+      if (this.base.file) {
+        console.log('[UPLOADING FILE]')
+        uploadDirect(this.base.file, this.accountId).catch(console.warn)
```
kkukelka, line 184: could we do the catch after `try` block? 
via
```
catch(err) {
...
}
```

components/unique/Create/CreateCollection.vue
```diff
@@ -215,11 +204,15 @@ export default class CreateCollection extends mixins(
...
-      //   ? mintString
-      //   : [this.toRemark(mintString), ...(await this.canSupport())]
+      const args = [
+        this.cretateArgs(randomId, metadata),
+        ...(await canSupport(this.hasSupport)),
+      ]
+
+      if (this.base.file) {
+        console.log('[UPLOADING FILE]')
+        uploadDirect(this.base.file, this.accountId).catch(console.warn)
```
kkukelka, line 214: here as well

kkukelka (approved):
lgtm, maybe 2 small adjustments?

components/rmrk/Create/CreateCollection.vue
vikiival: Yeah that’s the catch. I do not care if its uploaded or not

PR #2292, opened 2022-02-08
Show available balance at NFTs for BUY

chandradot99:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).
👇 \_ Let's make a quick check before the contribution.

### PR type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

### What's new?

- [x] PR closes #2282 
- [x] Added balance info and disabled buy button in case of insufficient balance.

### Before submitting Pull Request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [ ] I've posted a screenshot of demonstrated change in this PR

### Optional

- [ ] I've tested it at </rmrk/collection/26902bc2f7c20c546a-1FVG7>
- [ ] I've tested PR on mobile and everything seems works
- [ ] I found edge cases
- [ ] I've written some unit tests 🧪

### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=CeBWgisygADqqxGk7ju8aMNEKz1K2WkxMKC3G8HsMnm119n)

### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

### Screenshot

- [ ] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

PR #2373, opened 2022-02-16
single subscribe

vikiival:
This feature is needed for #2292 

The base is that when you component need to flexibly subscribe to single subscription with the flexible arguments

- :zap: added singleSubscribe Mixin
- :bug: linter fails if you have newer node than 16.13.2

### PR type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

### DIFF
--- a/utils/mixins/singleSubscribeMixin.ts
+++ b/utils/mixins/singleSubscribeMixin.ts
@@ -1,0 +1,28 @@
+import { Component, Vue } from 'nuxt-property-decorator'
+// declare type UnsubscribePromise = Promise<Unsubscribe>;
+declare type Unsubscribe = () => void
+
+/*
+ * refer to https://stackoverflow.com/questions/51873087/unable-to-use-mixins-in-vue-with-typescript
+ * usage import Component, { mixins } from 'vue-class-component';
+ * class ExtendedClass extends mixins(SubscribeMixin) {
+ */
+@Component
+export default class SingleSubscribeMixin extends Vue {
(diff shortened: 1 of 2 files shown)


chandradot99:
@vikiival I Mixins can be used in Vue component, but in our case of #2292, we are subscribing either from plugin or from Vuex Store, so I think we can use this Mixin there.

utils/mixins/singleSubscribeMixin.ts
```diff
@@ -0,0 +1,28 @@
...
+declare type Unsubscribe = () => void
+
+/*
+ * refer to https://stackoverflow.com/questions/51873087/unable-to-use-mixins-in-vue-with-typescript
+ * usage import Component, { mixins } from 'vue-class-component';
+ * class ExtendedClass extends mixins(SubscribeMixin) {
+ */
+@Component
+export default class SingleSubscribeMixin extends Vue {
+  private sub: Unsubscribe = () => void 0
```
roiLeo, line 12: `void 0`?

roiLeo (commented):
I don't really understand why and what this new mixins is for, but the code looks good to me.

vikiival:
Like the problem with plugin is that you can't unsubscribe

utils/mixins/singleSubscribeMixin.ts
vikiival: > why and what this new mixin is for

The use-case is that @chandradot99 has to subscribe for the user's balance.
We also need to correctly handle when the user logs in under different address

PR #2419, opened 2022-02-21
⚒️ landing prefix

roiLeo:
### PR type

- [x] Bugfix

### What's new?

- [x] hide sliders if prefix !== 'rmrk'


yangwao:
oh I guess we will tackle this smarter )

PR #2742, opened 2022-04-02
Implement sorting by date under holders address

Jarsen136:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).
👇 \_ Let's make a quick check before the contribution.

### PR type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

### What's new?

- [x] PR closes #2741
- [ ] <brief_description_of_what_I've_added>

### Before submitting Pull Request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [x] I've posted a screenshot of demonstrated change in this PR

### Optional

- [x] I've tested it at </rmrk/collection/26902bc2f7c20c546a-1FVG7>
- [x] I've tested PR on mobile and everything seems works
- [ ] I found edge cases
- [ ] I've written some unit tests 🧪

### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=Caiv9TbPz68q5dC8EcHu5xKYPRnremimGzqmEejDFNpWWLG)

### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

### Screenshot

- [x] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

image

PR #2725, opened 2022-04-01
fix: profile tabs not working properly sometimes

Jarsen136:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).
👇 \_ Let's make a quick check before the contribution.

### PR type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

### What's new?

- [x] PR closes #2723 
- [ ] <brief_description_of_what_I've_added>

### Before submitting Pull Request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [x] I've posted a screenshot of demonstrated change in this PR

### Optional

- [ ] I've tested it at </rmrk/collection/26902bc2f7c20c546a-1FVG7>
- [ ] I've tested PR on mobile and everything seems works
- [ ] I found edge cases
- [ ] I've written some unit tests 🧪

### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=Caiv9TbPz68q5dC8EcHu5xKYPRnremimGzqmEejDFNpWWLG)

### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

### Screenshot

- [ ] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

PR #2751, opened 2022-04-03
feat add holdings by user

### LINKED ISSUE #2739: Holdings by User
Seems KodaDot NFT marketplace is getting popular and taking lead in the best analytics, we would like to help users better scope on users activity.

In a similar manner, we have right now merging #2707  #2724 would be superior bringing a similar view with expanding custom rows and adding another tab "Holdings".
It would show a table. 
- Headers would be the name of Collection, Amount, Bought, Sale, Last Activity 
- Link Collection name to collections view
- link items in the collection to items 
- last activity in relative time with on hover with precise time

Best to reuse components if it's possible! 

Retrospect realized, we already have a very similar issue but from creator's perspective
- #1285 

Holdings by the user should be for collectors to show off, however, might be somehow merged in future.

image

ht ShawnLin for the suggestion

Jarsen136:

### PR type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

### What's new?

- [x] PR closes #2739
- [ ] <brief_description_of_what_I've_added>

### Before submitting Pull Request, please make sure:

### Optional

- [ ] I've tested it at </rmrk/collection/26902bc2f7c20c546a-1FVG7>
- [ ] I've tested PR on mobile and everything seems works
- [ ] I found edge cases
- [ ] I've written some unit tests 🧪

### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=Caiv9TbPz68q5dC8EcHu5xKYPRnremimGzqmEejDFNpWWLG)

### Community participation

### Screenshot

image


yangwao:
can't see https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo it redirect me to collected... 
but maybe it's fixed in #2725 ?

yangwao:
> can't see [deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo](https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo) it redirect me to collected...
> 
> same for [deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/JCHzJjQAbgk96DsGq69JgwPLXnc3Ab6UfGjtxYZ5eLV2vFW?tab=collected](https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/JCHzJjQAbgk96DsGq69JgwPLXnc3Ab6UfGjtxYZ5eLV2vFW?tab=collected)
> 
> I bet it has something to do if the user who did not create, won't see holdings?
> 
> here it works tho [deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/Cp9r5r9StL5VWQpqKXZiiWJQyPsehPjnkcdA1h2ewrSCbwo?tab=collected](https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/Cp9r5r9StL5VWQpqKXZiiWJQyPsehPjnkcdA1h2ewrSCbwo?tab=collected)
> 
> image
> 
> but maybe it's fixed in #2725 ?

Seems doesn't work 🥹  

https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo?tab=holdings

yangwao:
Otherwise is seems works nice https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FAZBj1G1H1Ls4nKn8ru7F31qku4Fhgzcce3D56hgTfJAj4Q?tab=holdings

just wanted inspect on some collectors tho! 
:)

Jarsen136:
> can't see https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo it redirect me to collected...
> 
> same for https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/JCHzJjQAbgk96DsGq69JgwPLXnc3Ab6UfGjtxYZ5eLV2vFW?tab=collected
> 
> I bet it has something to do if the user who did not create, won't see holdings?
> 


I'm not sure if it's because the user has a list that doesn't allow access to purchases list.  The interface for requesting collections list returned empty.  And it well for for other addresses.
image



> 
> but maybe it's fixed in #2725 ?

#2725 solves an occasional problem of displaying empty lists. Not the same problem as mentioned above

yangwao:
> I'm not sure if it's because the user is set to not allow access to the profile page.

I guess we don't have any setting like that 🤔

yangwao:
wait https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/CuHWHNcBt3ASMVSJmcJyiBWGxxiWLyjYoYbGjfhL4ovoeSd?tab=holdings&page=1

just noticed it shows only NFTs which were minted by my address?

Jarsen136:
> wait https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/CuHWHNcBt3ASMVSJmcJyiBWGxxiWLyjYoYbGjfhL4ovoeSd?tab=holdings&page=1
> 
> just noticed it shows only NFTs which were minted by my address?

Yup, now I have fixed it and it can show all nfts owned by users.

Jarsen136:
However, I have to use two queries to get all holding information, including `nftListOwned`(just added) and `allCollectionSaleEvents`.  On addresses with a large number of NFTs, performance will be poor and may cause a server-side error.

The best solution is to combine them into one request.
Specifically, it is to enable the `nftListOwned` interface to return the result of the `allCollectionSaleEvents` interface.
Maybe need some help from @vikiival @roiLeo 

```
query nftListOwned(
  $account: String!,
  $first: Int,
  $offset: Int,
  $orderBy: NftEntitiesOrderBy = BLOCK_NUMBER_DESC
) {
  nFTEntities(
    filter: {
      currentOwner: { equalTo: $account }
      name: { notLike: "%Kanaria%" }
      burned: { distinctFrom: true }
    }
    orderBy: [$orderBy, BLOCK_NUMBER_DESC]
    first: $first
    offset: $offset
  ) {
    totalCount
    nodes {
      ...nft
    }
  }
}
```
```
query allCollectionSaleEvents($ids: [ID!], $and: [EventWhereInput!]) {
  events(where: { nft: { collection: { id_in: $ids } }, AND: $and }) {
    meta
    interaction
    blockNumber
    timestamp
    caller
    id
    nft {
      id
      name
      collection {
        id
        name
      }
    }
  }
}
```

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -189,29 +196,36 @@ type TableRow = {
 @Component({ components })
 export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
   @Prop({ type: Array }) public events!: Interaction[]
-  @Prop({ type: Boolean, default: true })
-  private readonly openOnDefault!: boolean
   @Prop({ type: Boolean, default: false }) hideCollapse!: boolean
+  @Prop({ type: String, default: '' }) groupKeyOption: string
+  @Prop({ type: String, default: 'Name' }) nameHeaderLabel: string
+  @Prop({ type: String, default: 'Date' }) dateHeaderLabel: string
+  @Prop({ type: String, default: '' }) collapseTitleOption: string
```
roiLeo, line 203:
```suggestion
  @Prop({ type: String, default: '' }) groupKeyOption!: string
  @Prop({ type: String, default: 'Name' }) nameHeaderLabel!: string
  @Prop({ type: String, default: 'Date' }) dateHeaderLabel!: string
  @Prop({ type: String, default: '' }) collapseTitleOption!: string
```

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +365,57 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
+
+  private getCustomRowFilter() {
+    if (this.groupKey === 'Holder') {
+      return (item) => item.Holder !== '-'
+    } else if (this.groupKey === 'CollectionId') {
+      return (item) => item.Holder === this.$route.params.id
+    }
+  }
+
+  private generatecustomGroups(itemRowList): TableRow[] {
```
roiLeo, line 384: missing `itemRowList` type

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +365,57 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
+    const customGroups: Record<string, TableRow> = {}
+    itemRowList.filter(this.getCustomRowFilter()).forEach((item) => {
```
roiLeo, line 386: missing `item` type

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +365,57 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
+      item = {
+        Bought: 0,
+        Sale: 0,
+        ...item,
+      }
+      const groupName = this.getGroupNameFromRow(item)
+      if (customGroups[groupName]) {
+        customGroups[groupName].Items.push(item)
```
roiLeo, line 394: Object is possibly 'undefined'

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +365,57 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
 
-    holderGroupsList.forEach((group) => {
+    customGroupsList.forEach((group) => {
       parsePriceForItem(group, this.decimals, this.unit)
       group['Amount'] = group['Items'].length
       group['Items'].forEach((item) => {
         parsePriceForItem(item, this.decimals, this.unit)
       })
       group['Items'] = group['Items'].sort((a, b) => b.SortKey - a.SortKey)
     })
```
roiLeo, line 417: Object is possibly 'undefined'

components/rmrk/Gallery/Holding.vue
```diff
@@ -0,0 +1,87 @@
...
+import { sortedEventByDate } from '~/utils/sorting'
+import nftListOwned from '@/queries/nftListOwned.graphql'
+
+const components = {
+  Holder: () => import('@/components/rmrk/Gallery/Holder/Holder.vue'),
+}
+
+@Component({ components })
+export default class Holding extends mixins(PrefixMixin) {
+  @Prop({ type: String, default: '' }) accountId: string
```
roiLeo, line 27: you can use `AuthMixin` to get `accountId`

components/rmrk/Gallery/Holding.vue
```diff
@@ -0,0 +1,87 @@
...
+        }
+      },
+      fetchPolicy: 'cache-and-network',
+    })
+  }
+
+  protected async handleNftListResult({ data }: any) {
+    if (data && data.nFTEntities) {
+      const collectionIdSet = new Set<string>()
+      data.nFTEntities.nodes.forEach((item) => {
```
roiLeo, line 53: missing `item` type

components/rmrk/Gallery/Holder/Holder.vue
Jarsen136: Thanks. I will fix all types problems.

components/rmrk/Gallery/Holding.vue
Jarsen136: the accountId comes from AuthMixin is for the current user.

On the User profile page, accountId would be different for different users.

yangwao:
> However, I have to use two queries including nftListOwned(just added) and allCollectionSaleEvents, to get all holding information. On addresses with a large number of NFTs, performance will be poor and it may cause a server-side error.

We can do special magic and cache some queries which loads superior fast, same technique we are using in spotlight and series-insight iirc, as it's huge query.

yangwao:
Oh, I see an issue with big queries now.

https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo?tab=holdings

image

Jarsen136:
> We can do special magic and cache some queries which loads superior fast, same technique we are using in spotlight and series-insight iirc, as it's huge query.

Yes, you are right. And I think the quickest solution is to enable the `nftListOwned` interface to return events for specific nfts.  It can obviously solve the problem of too large response data.

vikiival:
> Specifically, it is to enable the nftListOwned interface to return the result of the allCollectionSaleEvents interface.

Not sure what do you mean.

Jarsen136:
> > Specifically, it is to enable the nftListOwned interface to return the result of the allCollectionSaleEvents interface.
> 
> Not sure what do you mean.

Hey, Let me expand more: 

Current process:
`nftListOwned` -> receive specific nfts list-> generate collectionId list  -> `allCollectionSaleEvents` -> receive all **events list of  specific collections**

What process I expected:
`nftListOwned` -> receive the **events list of specific nfts**

vikiival:
https://github.com/kodadot/nft-gallery/issues/2739#issuecomment-1087341498

vikiival:
feel free to test [here](https://app.gc.subsquid.io/beta/rubick/005/graphql)

query variables: 
```json
{
  "id": "Fksmad33PFxhrQXNYPPJozgWrv82zuFLvXK7Rh8m1xQhe98"
}
```

In case of last actitivity you can extend the query with  

```graphql
      lastUpdate: events(orderBy: timestamp_DESC, limit: 1) {
        id
        interaction
        blockNumber
        timestamp
        meta
      }
    }
```

Jarsen136:
> Here you go sir
> 
> ```graphql
> query collectionHoldingsByAccount($id: String!) {
>   collectionEntities(where: {nfts_some: {currentOwner_eq: $id}}) {
>     id
>     name
>     nfts(where: {currentOwner_eq: $id}) {
>       id
>       price
>       updatedAt
>       bought: events(where: {caller_eq: $id, interaction_eq: BUY}, orderBy: timestamp_DESC, limit: 1) {
>         id
>         meta
>       }
>     }
>   }
> }
> ```

Looks perfect! I will have a try. Though I have to do a big rework if I use this new query : )

Jarsen136:
Oh, I found a bad edge case that this query may unable to handle. Please take a look. @vikiival 

I need to know the latest time of these three events: SEND、BUY、MINT, so that I can sort the nfts correctly in a group.
The 'lastUpdate' does not meet the needs mostly.
ref: https://github.com/kodadot/nft-gallery/pull/2742#issuecomment-1086489051


BTW, Is it possible to use the solution I mentioned above, so that I can largely reuse the original components 👀

ma-shulgin:
> ### Testcase
> https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo?tab=holdings
> 
> Oh, I see an issue with big queries now. Seems giving me 504 which is gateway timeout, I'll ping the Subsquid squad so let's see how they can help us on this track
> 
> image
> 
> image

Hey! @Jarsen136 @yangwao will you display all thousands of event on the single page?😃 Try to paginate your queries. This will fix your problem. Use limit-offset into your events or eventsConnection

Jarsen136:
> Hey! @Jarsen136 @yangwao will you display all thousands of event on the single page?😃 Try to paginate your queries. This will fix your problem. Use limit-offset into your events or eventsConnection . Smth like this
> 

What you mentioned is a possible way, but it still needs to request many times to get the complete events list, because part of the display information needs to be analyzed based on the complete event.

It is not correct to display only paginated data.

ma-shulgin:
> > Hey! @Jarsen136 @yangwao will you display all thousands of event on the single page?😃 Try to paginate your queries. This will fix your problem. Use limit-offset into your events or eventsConnection . Smth like this
> 
> What you mentioned is a possible way, but it still needs to request many times to get the complete events list, because part of the display information needs to be analyzed based on the complete event.
> 
> It is not correct to display only paginated data.

But getting more than 100 000 objects via just one request is not an option too. So try to use connection

Jarsen136:
> > > Hey! @Jarsen136 @yangwao will you display all thousands of event on the single page?😃 Try to paginate your queries. This will fix your problem. Use limit-offset into your events or eventsConnection . Smth like this
> > 
> > 
> > What you mentioned is a possible way, but it still needs to request many times to get the complete events list, because part of the display information needs to be analyzed based on the complete event.
> > It is not correct to display only paginated data.
> 
> But getting more than 100 000 objects via just one request is not an option too. So try to use [connection](https://relay.dev/graphql/connections.htm)

Yup, pagination is important.  And I wonder if it is possible we request only partial nft events, which can significantly reduce the amount of data requested. 👀 
https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087326441

ma-shulgin:
> Imho best solution is to do it via collections and paginate it :) [#2751 (comment)](https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087352921)
> 
> cc @ma-shulgin

Yeah. This will speed up the response itself, but it's not enough to get all the events. So yeah you will need to paginate it too, and then spend resources to restruct it to the list of events. But such "collection-based" structure is harder to properly paginate. At last , if you will break it at least by 1000 events, you won't win a lot of time by getting all of them

ma-shulgin:
> > > > Hey! @Jarsen136 @yangwao will you display all thousands of event on the single page?😃 Try to paginate your queries. This will fix your problem. Use limit-offset into your events or eventsConnection . Smth like this
> > > 
> > > 
> > > What you mentioned is a possible way, but it still needs to request many times to get the complete events list, because part of the display information needs to be analyzed based on the complete event.
> > > It is not correct to display only paginated data.
> > 
> > 
> > But getting more than 100 000 objects via just one request is not an option too. So try to use [connection](https://relay.dev/graphql/connections.htm)
> 
> Yup, pagination is important. And I wonder if it is possible we request only partial nft events, which can significantly reduce the amount of data requested. 👀 [#2751 (comment)](https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087326441)

What do you mean? Filter them by some condition?

ma-shulgin:
@Jarsen136 And if it is not a secret, what kind of analytics you plan to do? May be we can add it directly to endpoint and cache it every several minutes?

Jarsen136:
> partial

Actually, we don't need the events of the entire collection, which is what `allCollectionSaleEvents` provide.
We only need the events associated with the nft owned by the user and it is enough.

Jarsen136:
> @Jarsen136 And if it is not a secret, what kind of analytics you plan to do? May be we can add it directly to endpoint and cache it every several minutes?

We need to know the time when the user owns an nft.
ref: https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087391969

ma-shulgin:
> > @Jarsen136 And if it is not a secret, what kind of analytics you plan to do? May be we can add it directly to endpoint and cache it every several minutes?
> 
> We need to know the time when the user owns an nft. ref: [#2751 (comment)](https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087391969)

When did current owner buy it or all buy events?

Jarsen136:
> > > @Jarsen136 And if it is not a secret, what kind of analytics you plan to do? May be we can add it directly to endpoint and cache it every several minutes?
> > 
> > 
> > We need to know the time when the user owns an nft. ref: [#2751 (comment)](https://github.com/kodadot/nft-gallery/pull/2751#issuecomment-1087391969)
> 
> When did current owner buy it or all buy events?

I mean the time when the current owner bought it, received it or minted it. 

for example,  

for this owner below, the time come from  buy event:
image


for this owner below, the time come from  mint event:
image

ma-shulgin:
@Jarsen136  if I understand you correctly this should help:

```graphql
query allCollectionSaleEvents($owner: String) {
  nftEntities(where: {currentOwner_eq: $owner}, limit: 100) {
    id
    name
    collection {
      name
      id
    }
    events(orderBy: timestamp_DESC, limit: 1, where: {interaction_in: [BUY, SEND, MINTNFT]}) {
      timestamp
      interaction
    }
  }
}

```

ma-shulgin:
> > @Jarsen136 if I understand you correctly this should help:
> > ```graphql
> > query whenOwnerGotThis($owner: String) {
> >   nftEntities(where: {currentOwner_eq: $owner}, limit: 100) {
> >     id
> >     name
> >     collection {
> >       name
> >       id
> >     }
> >     events(orderBy: timestamp_DESC, limit: 1, where: {interaction_in: [BUY, SEND, MINTNFT]}) {
> >       timestamp
> >       interaction
> >     }
> >   }
> > }
> > ```
> 
> Thanks very much for your idea. I was inspired by your query and wrote the query I needed! It seems to work well!
> 
> ```graphql
> query allCollectionSaleEventsByAccountId($id: String) {
>   nftEntities(where: {currentOwner_eq: $id}) {
>     id
>     events(orderBy: timestamp_DESC) {
>       meta
>       interaction
>       blockNumber
>       timestamp
>       caller
>       id
>       nft {
>         id
>         name
>         collection {
>           id
>           name
>         }
>       }
>     }
>   }
> }
> ```
> 
> image

Hey! Actually you don't need to get an nft from the event it will create another query inside db and slows down the response . You can get all the info from the initial nft entity from which you get events. So this will do exactly the same but will work faster.
```graphql
query allCollectionSaleEventsByAccountId($id: String) {
  nftEntities(where: {currentOwner_eq: $id}) {
    id
    name
    collection {
    	id
      name
    }
    events(orderBy: timestamp_DESC) {
      meta
      interaction
      blockNumber
      timestamp
      caller
      id
    }
  }
}
```

Jarsen136:
>can't see https://deploy-preview-2751--koda->nuxt.netlify.app/rmrk/u/FXUxw9Xp98gTzrU2z63D2G92kcx1oznd6xYa8WdqRgTcEDo it redirect me to collected...

>same for https://deploy-preview-2751--koda->nuxt.netlify.app/rmrk/u/JCHzJjQAbgk96DsGq69JgwPLXnc3Ab6UfGjtxYZ5eLV2vFW?tab=collected
> 

They are all working well now 🎉  @yangwao

Jarsen136:
> 
> Actually you don't need to get an nft from the event as it will create another query inside db and slows down the response . You can get all the info from the initial nft entity from which you get events. So this will do exactly the same but will work faster.
> 
> ```graphql
> query allCollectionSaleEventsByAccountId($id: String) {
>   nftEntities(where: {currentOwner_eq: $id}) {
>     id
>     name
>     collection {
>     	id
>       name
>     }
>     events(orderBy: timestamp_DESC) {
>       meta
>       interaction
>       blockNumber
>       timestamp
>       caller
>       id
>     }
>   }
> }
> ```

Yes, you are right. I will try to refactor it.

yangwao:
Ready to review?

Jarsen136:
> 

Sure it is ready

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +366,61 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
+  }
+
+  private generatecustomGroups(itemRowList: TableRow[]): TableRow[] {
+    const customGroups: Record<string, TableRow> = {}
+    itemRowList.filter(this.getCustomRowFilter()).forEach((item: TableRow) => {
+      item = {
+        Bought: 0,
+        Sale: 0,
+        ...item,
+      }
```
roiLeo, line 394: Ts error here, this is specified more than once + Sale is define as a string but you push number you should change type and do something like:
```ts
item.Bought = item.Sale = 0
```
not sure if this would be a good pratice

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +366,61 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
-            ...item,
-            Items: [item],
-          }
+  private getGroupNameFromRow(item: TableRow): string {
+    if (this.groupKey === 'Holder') {
+      return item['Holder']
+    } else if (this.groupKey === 'CollectionId') {
+      return item['Item']['collection']['id']
+    }
+    return item['Holder']
```
roiLeo, line 375: small refacto

```suggestion
		return this.groupKey !== 'Holder'
			? item['Item']['collection']['id']
			: item['Holder']
```

components/rmrk/Gallery/Holder/Holder.vue
```diff
@@ -345,37 +366,61 @@ export default class Holder extends mixins(ChainMixin, KeyboardEventsMixin) {
...
+  }
+
+  private getCustomRowFilter(): (item: TableRow) => boolean {
+    if (this.groupKey === 'Holder') {
+      return (item) => item.Holder !== '-'
+    } else if (this.groupKey === 'CollectionId') {
+      return (item) => item.Holder === this.$route.params.id
+    }
+    return () => true
```
roiLeo, line 384: another small refacto
```suggestion
    return this.groupKey !== 'Holder'
			? (item) => item.Holder === this.$route.params.id
			: (item) => item.Holder !== '-'
```

components/rmrk/Gallery/Holding.vue
```diff
@@ -0,0 +1,67 @@
...
+import { notificationTypes, showNotification } from '@/utils/notification'
+import { sortedEventByDate } from '~/utils/sorting'
+
+const components = {
+  Holder: () => import('@/components/rmrk/Gallery/Holder/Holder.vue'),
+}
+
+@Component({ components })
+export default class Holding extends mixins(PrefixMixin) {
+  @Prop({ type: String, default: '' }) accountId: string
```
roiLeo, line 26: no initializer and is not definitely assigned in the constructor

components/rmrk/Gallery/Holding.vue
```diff
@@ -0,0 +1,67 @@
...
+          const nftEvents = item.events
+          nftEvents.forEach((e) => {
+            e.nft = {
+              id: item.id,
+              name: item.name,
+              collection: item.collection,
+            }
+          })
+          events.push(...nftEvents)
+        })
```
roiLeo, line 52: Property 'nftEntities' does not exist on type '{ events: Interaction[]; }'
I would like to see `map` instead of nested `forEach`

components/rmrk/Gallery/Holding.vue
```diff
@@ -0,0 +1,67 @@
...
+  protected async fetchNftEvents() {
+    try {
+      const { data } = await this.$apollo.query<{ events: Interaction[] }>({
+        query: allNftSaleEventsByAccountId,
+        client: 'subsquid',
+        variables: {
+          id: this.accountId,
+        },
+      })
+      if (data && data.nftEntities && data.nftEntities.length) {
```
roiLeo, line 40: Property 'nftEntities' does not exist on type '{ events: Interaction[]; }'

queries/nftListOwned.graphql
```diff
@@ -0,0 +1,24 @@
+#import "./fragments/nft.graphql"
```
roiLeo, line 1: I don't think this file is used anywhere

queries/rmrk/subsquid/allNftSaleEventsByAccountId.graphql
```diff
@@ -0,0 +1,18 @@
+query allNftSaleEventsByAccountId($id: String) {
+  nftEntities(where: {currentOwner_eq: $id}) {
```
roiLeo, line 2: I would filter burned nft `{ burned_eq: false }`

yangwao:
A hovercard could be added as well as now it doesn't work :|

Can make a separate issue or make it here with extra bounty, up to you.

image

Jarsen136:
> A hovercard could be added as well as now it doesn't work :|
> 
> 
> 
> Can make a separate issue or make it here with extra bounty, up to you.
> 
> 
> 
> image
> 
> 

okey, i will make it here.

yangwao:
Oh, I've just noticed by using it, 
would be nice if when I 
- go from **collection holders tab**
- Click on the user in the table 
- would redirect me to the holdings tab of the user directly

Small tweak which makes contextual sense :)

components/rmrk/Gallery/Holder/Holder.vue
Jarsen136: It seems that it don't need to change, because we may add more group type in the future, not just 'Holder' and 'CollectionId'.

components/rmrk/Gallery/Holder/Holder.vue
Jarsen136: It seems that it don't need to change, because we may add more group type in the future, not just 'Holder' and 'CollectionId'.

components/rmrk/Gallery/Holder/Holder.vue
roiLeo: Alright, in that case, for later, I would suggest to use a switch case or an object lookups to avoid having a long if...else statements

components/rmrk/Profile/ProfileDetail.vue
```diff
@@ -381,7 +392,7 @@ export default class Profile extends mixins(PrefixMixin) {
       this.collections = data.collectionEntities
     }
     // in case user is only a collector, set tab to collected
-    if (this.totalCollections === 0) {
+    if (this.totalCollections === 0 && this.$router.query.tab !== 'holdings') {
```
roiLeo, line 395:
```suggestion
    if (this.totalCollections === 0 && this.$route.query.tab !== 'holdings') {
```

Jarsen136: you are very correct

components/rmrk/Gallery/Holder/Holder.vue
roiLeo: @Jarsen136 these lines are still problematic, I've seen that you fixed `Sale` type but I still got an error here

Jarsen136: That is really confusing. Could you please give me the report url, so that I can check where the error come from : )

components/rmrk/Gallery/Holding.vue
Jarsen136: will fix later.

components/rmrk/Gallery/Holder/Holder.vue
roiLeo: weird fix but ok

Jarsen136: : ) because the zero value is a fallback value, could not be set to Sate\Bought directly, which will cause an incorrect result.

yangwao:
Can we cut long name 👀

https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/GAYBQX5f7NXS4dv5EnL7NRSGLnAQmoDqajx998nYLv8QoJT?tab=holdings&page=3

🥹  

image

Jarsen136:
> Can we cut long name 👀
> 
> https://deploy-preview-2751--koda-nuxt.netlify.app/rmrk/u/GAYBQX5f7NXS4dv5EnL7NRSGLnAQmoDqajx998nYLv8QoJT?tab=holdings&page=3
> 
> 🥹
> 
> image

Sure, a piece of cake.

components/rmrk/Gallery/Holding.vue
Jarsen136: Is it resolved now?

components/rmrk/Gallery/Holding.vue
Jarsen136: Is it resolved now? 👀

roiLeo: `...nftEvents` is in error: "Argument of type 'any' is not assignable to parameter of type 'never'"

roiLeo: hotfix could be `const events: any[] = []`

roiLeo (approved):
code is error free, we might need to replace all `any` type later

Jarsen136:
> code is error free, we might need to replace all `any` type later

Ok,  I will replace these 'any' type later in this issue #2785

yangwao:
pay 300 usd

yangwao:
interesting lol 
Screenshot 2022-04-06 at 17 04 48

Issue #3334, opened 2022-07-05
Rework current pinning & storage strategy


yangwao:
Hence there were unknown changes or at least on Basilisk we've noticed that after fresh minting things are not working. 
Thus time to revisit the initial implementation of cf workers and how we have it in the current state.

Currently, the user pins stuff on Estuary, which we've found acting suboptimal as it's being stored on Filecoin (archive nodes(?) and retrieving through pinata gateway takes a long time (peering issues?). (Something has changed outside of our control, we are investigating, it till we are looking to mitigate it and make a more resilient and antifragile scenario)    

### Future proposed changes are 
- [x] #3358 temp hotfix
- [ ] Pinning metadata on Pinata (as we are blazingly fast here for end-user experience) and rest pinning could be kept on Estuary -> We are pinning rn on NFT.storage and seems temporary hotfix is okay
- [ ] At minting, we should push media to CF images (might not happen during BSX mints?)
- [ ] At the new implementation, we would doesn't need durable objects at all and fetch images by IPFS hash from CF images to accelerate end-user experience while they browsing collections/explorer
- [x] https://github.com/kodadot/nft-gallery/issues/3745
- [x] https://github.com/kodadot/nft-gallery/issues/3296

### Long-term goal

Currently, we are fairly doing close to 1TB on Pinata Gateway monthly as transfers for users and this would be an interesting way to offload as public goods for storage providers, like those around the Filecoin provider's circle

### Suggested would be to recycle code from
- https://github.com/kodadot/nft-gallery/blob/main/src-functions/pinJson.js 
- https://github.com/kodadot/workers/blob/master/pinning/src/lib.rs#L84

To make it work with Pinata directly from workers to pin JSON metadata there

Check the links below for more context 
Let's comment below if you want to thrill on this issue 

### Ref
- https://github.com/kodadot/packages/issues/17
- https://github.com/kodadot/workers/issues/7
- https://github.com/kodadot/workers/issues/8
- https://github.com/kodadot/nft-gallery/issues/3253
- https://github.com/kodadot/bao/issues/10

yangwao:
> we are fairly doing close to 1TB on Pinata Gateway monthly

well, it was over 2TB recently - Total Requests: 4049068 - Bandwidth: 1.82 TB

yangwao:
Hence we have 
- #4587 

I'll review this one, but there is not much to do. (?)

yangwao:
Happy solved

Issue #3310, opened 2022-06-30
Refactor: create service folder


petersopko:
following from @roiLeo comment here:

- https://github.com/kodadot/nft-gallery/discussions/3068#discussioncomment-2855995

Here's the original comment: 

- create a service folder in order to organise all the actions (vuex actions -> service call)

I'd like to ask @roiLeo to provide additional information here if possible, I'm out of my depth with regards to services

roiLeo:
`services`

vikiival:
is this stil valid @kodadot/internal-dev

Issue #3291, opened 2022-06-28
Freshly minted collection doesn't show up


petersopko:
After creating new collection, I'm being redirected to https://beta.kodadot.xyz/bsx/create where I can mint new NFT, however, I don't see the created collection in the dropdown select. After some time and after I refresh the page, collection ID shows up in the dropdown (number) but not the name of the collection. 
![image](https://user-images.githubusercontent.com/73316633/176089723-69ae62e7-b56a-44f9-acb5-77915ad26054.png)

2 problems here: 
-  It's possible this will always take time regardless, but if that's the case, I'd suggest showing up message which will let the user know, to refresh in 5 mins or so. 
- 10 minutes after the mint, I can still see only the ID of the collection (number) not the name I've selected before

yangwao:
The collection name is being shown up already, yet it still takes some time for a fresh collection is being shown.

yangwao:
Explainer will be added 
- https://github.com/kodadot/nft-gallery/issues/3333

The fix could be adding 
- fetching for a user recent new collections through indexer as right now we are doing it on a fresh load 
- fetch from the chain by address recent collections from last blocks by user-id

petersopko:
ok, the explainer was added and this also doesn't seem to be an issue after I've tested this a bunch of times in the past couple of days, idk whether you want to move forward with:
> The fix could be adding

> fetching for a user recent new collections through indexer as right now we are doing it on a fresh load
> fetch from the chain by address recent collections from last blocks by user-id

But in case yes, maybe we should include this as part of our refactor, wdyt @yangwao 

![image](https://user-images.githubusercontent.com/73316633/183417485-98160513-7049-40ff-9007-f8d5d1d411db.png)

petersopko:
just tested this couple of times, we have both the message included and also, I just saw the collection show up instantly.

PR #3358, opened 2022-07-06
NFT Storage is here

vikiival:
- :zap: added nft-storage worker
- :zap: implemented nft-storage
- :zap: use NFT-Storage for Basilisk
- :zap: should refresh token if user had legacy ESTUARY one
- :zap: RMRK is using Nft storage
- :zap: unique is using nft storage
- :zap: app is doing direct upload

### PR type

- [x] Bugfix
- [x] Feature
- [x] Refactoring

### What's new?

- [ ] PR closes #<issue_number>
- [ ] Requires deployment <rubick/magick_issue>
- [ ] <brief_description_of_what_I've_added>

### Before submitting Pull Request, please make sure:


utils/nftStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    const { status, data } = await api.post<StorageApiResponse>(
+      '/pinFile',
+      file,
+      {
+        headers: {
+          'Content-Type': file.type ? `${file.type}` : '*/*',
+        },
+      }
+    )
+    console.log('[NFT::STORAGE] Pin Image', status)
```
roiLeo, line 67:
```suggestion
    console.log('[NFT::STORAGE] Pin Image', status)
```

could we use consola 🥺 
+
phrasing
`[NFT::STORAGE] Pin File`

roiLeo (commented):
`nftStorage` definitely looks like something that can be moved in services folder (ref #3310)

By the way I find something strange, `$nuxt.context` should be accessible from this file but it's not, do you have any idea why?

yangwao:
right away seems it works well yet, immediately returns 500 but reloading right away wfm 

Screenshot 2022-07-07 at 10 53 40

yangwao:
now it worked that returned metadata but no picture even after reload and w/o error, strange

image

yangwao:
lmao, one can't simply think of this  

image

image

yangwao:
Seems right away now it has worked, but doubt it could be the case every time. 

I would suggest it fetch it through pinata on the background one more time while minting has finished. 

And delay user going to fresh mint to like 5 seconds 

image

petersopko:
minted collection and it's visible with name right away, seems to be closing 
- https://github.com/kodadot/nft-gallery/issues/3291
image

redirected to minting NFT after which minted NFT shows up right away:
image

https://deploy-preview-3358--koda-nuxt.netlify.app/bsx/gallery/2600576003-1

utils/nftStorage.ts
yangwao: #3310 ?

yangwao:
lgtm, works right away after mint. 
has someone tested it for rmrk?

vikiival:
> has someone tested it for rmrk?

[Magick](https://kodadot.xyz/rmrk/gallery/13445851-8cc1b91e899d9be40e-2CENTS-MAGICK-0000000000000005)

yangwao:
The only downside at the current implementation we are limited to 100MB per file until we rework it so that clients push it directly to NFT storage endpoint
- #3334 

image

PR #4298, opened 2022-11-10
:snake: Stuck Snek

vikiival:
# This is a temporary fix, please revert when snek goes to normal

- :ambulance: disable bsx minting
- :ambulance: disable bsx actions


yangwao:
snek will come back at approximately in 5-6 days once unbrick referendum will be over 

https://jakpan.hashnode.dev/snek-stall-post-mortem

PR #4476, opened 2022-12-11
fix: #4302 add wallet name

petersopko:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).

👇 \_ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #4302, Closes #4126

#### Before submitting pull request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [x] I've posted a screenshot of demonstrated change in this PR

#### Optional

- [x] I've tested it at </bsx/collection>
- [x] I've tested PR on mobile
- [ ] I've written unit tests 🧪
- [ ] I've found edge cases

#### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=FNyt7T1xdbhN7dagf7yGYCRJuE3R45VmTCE3tjzy1rKxa7y)

#### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

## Screenshot 📸

- [x] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

https://user-images.githubusercontent.com/73316633/206902254-9b400645-7642-436a-b3e0-7b34f1216594.mp4

PR #4486, opened 2022-12-13
🔧 en translation keyword

roiLeo:
continue https://github.com/kodadot/nft-gallery/pull/4476#discussion_r1047279041


yangwao:
quick pr before pushing to `hyper-jpeg` be like

PR #4624, opened 2023-01-06
🐛 responsivity by width (not touch)

### LINKED ISSUE #4622: Landing page tablet view
IMG_0027
- the navbar links have chevrons. Should look like the same as on the desktop.

petersopko:

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Closes #4622
- [ ] Requires deployment <>

#### Optional

- [x] I've tested it at </bsx/collection>
- [x] I've tested PR on mobile
- [ ] I've found edge cases

#### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=FNyt7T1xdbhN7dagf7yGYCRJuE3R45VmTCE3tjzy1rKxa7y)

#### Community participation

## Screenshot 📸

image


petersopko:
The issue was that we were relying on the touch aspect of the device instead of the resolution. 

The remaining issue is, that if the navbar on the tablet should be the same as on the desktop, the `Explore` button will always lead the user to the explore part of the app ->  there's no hover on the tablet (Also, the requirement from @exezbcz was to have click/tap on `Explore` to lead to the explore part of the app).

Should we disable the rerouting from the navbar to `Explore` on mobile devices?

exezbcz:
> The issue was that we were relying on the touch aspect of the device instead of the resolution.
> 
> The remaining issue is that if the navbar on the tablet should be the same as on the desktop, the `Explore` button will always lead the user to the explore part of the app -> there's no hover on the tablet (Also, the requirement from @exezbcz was to have click/tap on `Explore` to lead to the explore part of the app).
> 
> Should we disable the rerouting from the navbar to `Explore` on mobile devices? // I did this; currently, the link to Explore by clicking/tapping on it works only on a desktop; let me know if you want something else.

yup, in case of tablet, there wont be any selection displayed, after clicking on explore, it will take you to explore page.

petersopko:
> > 
> 
> the dropdown still shows for a sec on tablet when clicking

hmm that's weird, can't reproduce it from devtools tablet but I guess there's a momentary hover state activated even on tablet, can be disabled tho.

followup question: what about the chevron? If the `Explore` button on small screen works only as link to explore section, shouldn't we hide it as well?

exezbcz:
> > > 
> > 
> > 
> > the dropdown still shows for a sec on tablet when clicking
> 
> hmm that's weird, can't reproduce it from devtools tablet but I guess there's a momentary hover state activated even on tablet, can be disabled tho.
> 
> followup question: what about the chevron? If the `Explore` button on small screen works only as link to explore section, shouldn't we hide it as well?

good q, if its possible and not that hard, i would hide it on touch devices and leave it only for devices that have option to hover.

petersopko:
> good q, if its possible and not that hard, i would hide it on touch devices and leave it only for devices that have option to hover.

🚀

styles/layouts/_main-navigation.scss
```diff
@@ -318,6 +318,10 @@
     }
   }
 
+  .navbar-create {
+    margin-left: 0 !important;
```
Jarsen136, line 322: how about use class name`ml-0`?

components/Navbar.vue
```diff
@@ -376,7 +375,7 @@ export default class NavbarMenu extends mixins(
     window.addEventListener('scroll', this.onScroll)
     document.body.style.overflowY = 'initial'
     window.addEventListener('resize', () => {
-      this.isMobile = window.innerWidth < 1024 ? true : isMobileDevice
+      this.isMobile = window.innerWidth < 1024
     })
```
Jarsen136, line 379: I don't see the removeEventListener below. could you pls add it?

components/Navbar.vue
```diff
@@ -219,7 +218,7 @@ export default class NavbarMenu extends mixins(
   private lastScrollPosition = 0
   private artistName = ''
   public isBurgerMenuOpened = false
-  private isMobile = window.innerWidth < 1024 ? true : isMobileDevice
```
Jarsen136, line 222: I look through the git history. This line seems to fix a bug on laptops. Will it introduce original bug after the change?
image

Jarsen136 (changes requested):

components/Navbar.vue
petersopko: added, I'd like to know if we need to remove the event listener for resizing if the resizing can happen anytime while using the app. (i.e., should we be listening to it all the time?) Navbar is part of the layout everywhere.

styles/layouts/_main-navigation.scss
petersopko: prettier, thank you.

components/Navbar.vue
petersopko: I don't know which bug you are referring to, but generally, we've moved from caring about the touch/"mobileness" of the device and moved to, strictly speaking, screen size differentiation.

Small screens are treated as mobiles when it comes to the navbar. I'm unsure if this is right from the UX perspective, but this was the requirement.

components/Navbar.vue
Jarsen136: Yes, we should listen `resize` all the time. the resize event only be triggered when we change the window size.

Jarsen136 (approved):
code lgtm

yangwao:
pay 30 usd

yangwao:
💵 $30 @ 24.85 USD/KSM ~ 1.207 $KSM
🧗 [FNyt7T1xdbhN7dagf7yGYCRJuE3R45VmTCE3tjzy1rKxa7y](https://kusama.subscan.io/account/FNyt7T1xdbhN7dagf7yGYCRJuE3R45VmTCE3tjzy1rKxa7y)
🔗 [0xfbd7ceab99536bb4c4b8de2bbf30ade58b0b233da6b9c79ab4bc78c7fe0aefe8](https://kusama.subscan.io/extrinsic/0xfbd7ceab99536bb4c4b8de2bbf30ade58b0b233da6b9c79ab4bc78c7fe0aefe8)

PR #4671, opened 2023-01-12
chore: enable source maps for sentry

preschian:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).
👇 \_ Let's make a quick check before the contribution.

### PR type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

### What's new?

- [x] PR closes #4670

### Before submitting Pull Request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [ ] I've posted a screenshot of demonstrated change in this PR

### Optional

- [ ] I've tested it at </rmrk/collection/26902bc2f7c20c546a-1FVG7>
- [ ] I've tested PR on mobile and everything seems works
- [ ] I found edge cases
- [ ] I've written some unit tests 🧪

### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=DY4SQF2iD456tH89aQtz5wv1EV3BbSW8wKKuMcwbmXaj1pM)

### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

### Screenshot

- [ ] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

PR #4680, opened 2023-01-13
fix: enable sourcemap if token exists

preschian:
previously https://github.com/kodadot/nft-gallery/pull/4671/files#diff-65d5d69bb59dcc81ad971719549cdee5018f0a240add8b7ef5c65bee3e73dae0R323-R324 I put the wrong logic. with this, forked repository doesn't need to add additional env in their github

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -320,8 +320,8 @@ export default defineNuxtConfig({
     ],
     extend(config) {
       if (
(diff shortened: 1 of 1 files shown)


roiLeo (approved):
Do we need to set `SENTRY_AUTH_TOKEN` in env to make it work in localhost?

PR #4758, opened 2023-01-18
feat: Mobile Gallery Item: Full width on buy & Make Offer

Jarsen136:
**Thank you for your contribution** to the [KodaDot NFT gallery](https://kodadot.xyz).

👇 \_ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #4742
- [ ] Requires deployment <>

#### Before submitting pull request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality
- [x] I've posted a screenshot of demonstrated change in this PR

#### Optional

- [ ] I've tested it at </bsx/collection>
- [ ] I've tested PR on mobile
- [ ] I've written unit tests 🧪
- [ ] I've found edge cases

#### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=Caiv9TbPz68q5dC8EcHu5xKYPRnremimGzqmEejDFNpWWLG)

#### Community participation

- [x] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

## Screenshot 📸

- [x] My fix has changed **something** on UI; a screenshot is best to understand changes for others.
image

image

PR #4778, opened 2023-01-19
fix: disable buy button

Jarsen136:

## PR Type

- [x] Bugfix

There was a case that wasn't solved in the original pr #4758. Now fixed. 

Disabled buy button:

image

image

### DIFF
--- a/components/gallery/GalleryItemAction/GalleryItemActionSlides.vue
+++ b/components/gallery/GalleryItemAction/GalleryItemActionSlides.vue
@@ -76,9 +76,10 @@ defineProps<{
   width: 10rem;
   font-size: 1rem;
   position: relative;
-  .wrapper {
+  .neo-tooltip {
     width: 100%;
     height: 100%;
+    position: absolute;
   }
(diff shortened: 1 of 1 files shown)


vikiival:
@exezbcz does it fix the issue?

exezbcz:
there is no tooltip now tho; not sure whether its because of this pr. There should be "not enough founds" when you have connected your wallet, once this is fixed i think its good to go.

Jarsen136:
> there is no tooltip now tho; not sure whether its because of this pr. There should be "not enough founds" when you have connected your wallet, once this is fixed i think its good to go.

Yes, it is also a leftover bug.

✅  Fixed
image
cc @exezbcz

roiLeo (approved):
✅ wfm

Capture d’écran 2023-01-20 à 8 25 54 AM

roiLeo:
> Uh oh, there's something wrong with your config.

giphy copie

yangwao:
pay 50 usd

PR #4791, opened 2023-01-21
refactor: unify `/detail` page to `pages/_prefix/detail/_id/index.vue`

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] related with https://github.com/kodadot/nft-gallery/issues/4744. remove some `idb-keyval` on `GalleryItem`
- [x] unify `/detail` page to support multi chains with new design


vikiival:
Oof this is like 3 separated PRs

vikiival:
Maybe I am missing something

@preschian why were Available actions removed?

components/rmrk/Gallery/History.vue
```diff
@@ -377,7 +377,8 @@ export default class History extends mixins(
       event['Block'] = String(newEvent['blockNumber'])
 
       // ID for b-table: Use a unique key of your data Object for each row.
-      event['ID'] = newEvent['timestamp'] + newEvent['id']
+      event['ID'] =
```
vikiival, line 380: In Squid Event id is
`export const eventId = (id: string, event: Action) => `${id}-${event}${nanoid()}``

vikiival (changes requested):
Need clarification on removal

pages/snek/detail/_id.vue
```diff
@@ -1,28 +0,0 @@
...
-  },
-  head() {
-    return {
-      link: [
-        {
-          hid: 'canonical',
-          rel: 'canonical',
-          href: this.$root.$config.public.baseUrl + this.$route.path,
-        },
-      ],
```
Jarsen136, line 23: I wonder how could we handle the head config then?

roiLeo: `useNuxt2Meta`? and after Nuxt3 migration [useHead](https://nuxt.com/docs/getting-started/seo-meta)

roiLeo (commented):
💣 time to nuke some pages, code lgtm

but we still have two times the same page:
- `_prefix/gallery/_id/index`
- `_prefix/detail/_id/index` 

Don't we need a fallback if `prefix` doesn't exist?

vikiival (approved):
Let it roll

vikiival:
pay 45 usd

Issue #5067, opened 2023-02-21
[E2E] Test for creating collection is broken


vikiival:
In fix #5037

I disabled create button when name and symbol are empty.
Therefore this test is failing.

It would be good to fix the test. 

```
(Attempt 1 of 4) should trigger create collection button on rmrk
[803](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:805)
  (Attempt 2 of 4) should trigger create collection button on rmrk
[804](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:806)
  (Attempt 3 of 4) should trigger create collection button on rmrk
[805](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:807)
  1) should trigger create collection button on rmrk
[806](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:808)

[807](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:809)
  2 passing (1m)
[808](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:810)
  1 failing
[809](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:811)

[810](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:812)
  1) should trigger create collection button on rmrk:
[811](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:813)
     CypressError: Timed out retrying after 10050ms: `cy.click()` failed because this element is `disabled`:
[812](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:814)

[813](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:815)
`<button type="button" disabled="disabled" class="button is-primary is-outlined is-fullwidth">...</button>`
[814](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:816)

[815](https://github.com/kodadot/nft-gallery/actions/runs/4227161554/jobs/7341351968#step:9:817)
Fix this problem, or use `{force: true}` to disable error checking.
[...]

PR #5070, opened 2023-02-21
🔧 disabled submit button

roiLeo:
No disabled pattern

## PR Type

- [x] Bugfix

## Context

- [x] ref #5067

### DIFF
--- a/components/rmrk/Create/CreateCollection.vue
+++ b/components/rmrk/Create/CreateCollection.vue
@@ -43,7 +43,6 @@
           :message="balanceNotEnoughMessage">
           <SubmitButton
             expanded
-            :disabled="disabled"
             label="create collection"
             :loading="isLoading"
             @click="submit" />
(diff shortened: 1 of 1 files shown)


vikiival:
Nooo it will break the ui 😢

roiLeo:
> Nooo it will break the ui 😢

What do you mean? I'm not able to submit when input are required.
Why would you introduce disabled button only on specific chain?

vikiival:
I made it disabled because i was able to make collection with empty metadata

> why only one chain?

I forgot to add disabled to other chains 🥺

roiLeo:
> I made it disabled because i was able to make collection with empty metadata

How? I can't reproduce empty form. 
It doesn't need to be that way instead wa can try to fix form.

vikiival:
So the correct fix should be on the side of a package 

- https://github.com/kodadot/packages/issues/105

vikiival (approved):

PR #5057, opened 2023-02-20
✨ migrate store to pinia

roiLeo:
## Context

- [x] ref #4755
- [x] wanted to try out pinia store, looks like it's working well
- [x] delete unused component :(
- [x] works with nuxt2 class api
- [x] migrate lang vuex store to pinia
- [x] migrate wallet vuex store to pinia

## TODO
- [ ] persist selected language in localstorage

PR #5081, opened 2023-02-21
✨ useLocalStorage composable

roiLeo:
Might be needed (#5057)

### Usage

```ts
const walletProvider = useLocalStorage('wallet')
```

### DIFF
--- a/composables/useLocalStorage.ts
+++ b/composables/useLocalStorage.ts
@@ -1,0 +1,17 @@
+export default function (key: string, defaultValue: unknown) {
+  return customRef((track, trigger) => ({
+    get: () => {
+      track()
+      const value = localStorage.getItem(key)
+      return value ? JSON.parse(value) : defaultValue
+    },
+    set: (value) => {
+      if (value === null) {
+        localStorage.removeItem(key)
+      } else {
+        localStorage.setItem(key, JSON.stringify(value))
+      }
+      trigger()
+    },
+  }))
+}


vikiival:
Wondering why not cone from vue-use?

roiLeo:
> Wondering why not cone from vue-use?

Forgot [useStorage](https://vueuse.org/core/useStorage/#usestorage) exist.

> When using with Nuxt 3, this functions will NOT be auto imported in favor of Nitro's built-in [useStorage()](https://nitro.unjs.io/guide/introduction/storage). Use explicit import if you want to use the function from VueUse.

I prefer to rely on simple tools rather than having to maintain package version upgrade.
Feel free to close it if not needed.

vikiival:
> > Wondering why not cone from vue-use?
> 
> 
> 
> Forgot [useStorage](https://vueuse.org/core/useStorage/#usestorage) exist.
> 
> 
> 
> > When using with Nuxt 3, this functions will NOT be auto imported in favor of Nitro's built-in [useStorage()](https://nitro.unjs.io/guide/introduction/storage). Use explicit import if you want to use the function from VueUse.
> 
> 
> 
> I prefer to rely on simple tools r - ather than having to maintain package version upgrade.
> 
> Feel free to close it if not needed.


If we reexport use storage to use local storage

vikiival:
cc @roiLeo

roiLeo:
> If we reexport use storage to use local storage

14608107_1180665285312703_1558693314_n

vikiival (approved):

Issue #4857, opened 2023-01-29
Smaller side margins


exezbcz:
We should make the margins on fullhd smaller. Its than taking too much space for example in explore. 
![image](https://user-images.githubusercontent.com/90852205/215323213-00342928-5da6-4544-8a97-e3ea09741ac9.png)
something like this everywhere would be good
![image](https://user-images.githubusercontent.com/90852205/215323244-ac2ac02e-8cc3-4ecf-b0c2-d3acae0659ad.png)
is it possible? 

there is set padding to 48px
![image](https://user-images.githubusercontent.com/90852205/215545914-4b394c03-cac9-4cb8-9f85-21b66a28d216.png)
but the content seem to not follow it 

I'm now using a 12 column grid with a 48px margin. On mobile there is 20px margin

exezbcz:
would be good to have this after
- #4826

exezbcz:
Is there a method to adjust this or does it apply responsively depending on the situation?
<img src="https://media0.giphy.com/media/3oz8xT2mfj0r9YpBwk/giphy.gif"/>

roiLeo:
> Is there a method to adjust this or does it apply responsively depending on the situation? 

Idk what you mean by that but we can override [size of container](https://bulma.io/documentation/layout/container/) and set personalized [spacing helpers](https://bulma.io/documentation/helpers/spacing-helpers/)

exezbcz:
![Group 227](https://user-images.githubusercontent.com/90852205/220167891-a8fb72e1-3b03-4dc0-acd6-e599bca80b9d.png)

my bad :D 

Yes, I think that is it
- sometimes the margin left right is good but sometimes its taking up too much space on screen

![image](https://user-images.githubusercontent.com/90852205/220167799-504b821b-91e1-4f6d-bfb2-d6183877f0b2.png)

Wdyt?

roiLeo:
I can't reproduce container overflowing, could you share what size of screen?
![Capture d’écran 2023-02-21 à 2 26 57 PM](https://user-images.githubusercontent.com/9987732/220357166-faa5e820-865c-41ac-b646-09ac582365f6.png)

Or maybe I'm wrong way, you're waiting for something else? Increase margin on larger screen? Reduce container margin? Reponsive adaptation?

exezbcz:
@roiLeo yy, that was not the object of this. 
I am talking about making the side margins of container smaller - context is that I think we should not waste space on explore and stretch the content bit more. 

[...]

PR #5089, opened 2023-02-22
[Feedback request] 🔧 container is-fluid

roiLeo:
This is for testing purposes only 🧪 
ping @exezbcz 

## Context

- [x] ref #4857
- [x] container size for Navbar
- [x] container size for /explore pages

## Preview

https://deploy-preview-5089--koda-nuxt.netlify.app/bsx/explore/collectibles


yangwao:
actually, when netlify first pick ups PR and you move it to draft, when new commit event is emitted, it builds it anyway even in draft...

vikiival:
Please make this as draft

vikiival:
👀 can I close?

roiLeo:
> 👀 can I close?

I'm waiting for @exezbcz feedback to push changes

exezbcz:
@roiLeo hmm, it's better but still its something missing 
- I am also not completely sure what it does, the margins are now flexible or? 
- I can see that on bigger screens the content is stretching now - There is no dynamic grid yet so the collection cards are for example stretched. 
- but for example on 1240px the margins are looking worse than on current beta.
- there seems to be another 32px padding added
image

the objective was to lower the margins on certain screens, or best, make it unified for desktop and mobile, mobile has fixed 20px afaik. 

do I understand this correctly? Correct me please if I am wrong :D

roiLeo:
> I am also not completely sure what it does, the margins are now flexible or?
> I can see that on bigger screens the content is stretching now 

we can set fixed margin (left & right) for container, on this preview it's set as 32px and content will fulfill container

> There is no dynamic grid yet so the collection cards are for example stretched.

not related but we can rework resizing of collections cards

> but for example on 1240px the margins are looking worse than on current beta.

Looks oki?
Capture d’écran 2023-03-06 à 4 49 33 PM

> there seems to be another 32px padding added

This will be default padding (can be changed) between the content and the edge of the window

exezbcz:
@roiLeo yup, that is better now. 
I think something like 40px margin for desktop would be good, and then 20px for specific lower breakpoints - 1024?

vikiival:
cc @exezbcz need your visual 👁️ 👁️ 👁️

styles/global.scss
```diff
@@ -101,6 +101,11 @@ body {
   }
 }
 
+.container.is-fluid {
+  padding-left: 40px;
```
vikiival, line 105: maybe use `rem` instead of `px`?

styles/global.scss
```diff
@@ -113,6 +118,10 @@ body {
     padding-right: 10px !important;
     margin: 0 -12px !important;
   }
+  .container.is-fluid {
```
vikiival, line 121: likewise

vikiival (changes requested):
Smol

exezbcz:
yup, its okay, but its not working on landing page - is that expected? 

and with bigger items on explore - but that would be fixed individually no? 
image

roiLeo:
> yup, its okay, but its not working on landing page - is that expected?

Now should be testable on landing page

> and with bigger items on explore - but that would be fixed individually no? 

yeah we could change fixed width to make it more responsive

exezbcz:
@roiLeo it definitely makes sense for collection page and for explore page 
- We should apply it to the rest so its consistent, wdyt?

roiLeo:
> We should apply it to the rest so its consistent, wdyt?

Yes! Note that we will keep some pages at half size layout (teleport, create..) for now

exezbcz:
@roiLeo yup, It makes no sense making teleport full width, maybe once it is portfolio page? 👀 

Thank you for the changes!

exezbcz:
looking good! We will need to change the carousel on landing but otherwise solid.

yangwao:
mergeable? would be nice to push it out for release

vikiival (approved):

vikiival:
pay 50 usd

roiLeo:
I didnt had Time to test it on a large screen :/. How does it look ?

PR #5120, opened 2023-02-26
Collection redesign under banner section

### LINKED ISSUE #5073: Bellow banner section
task details in
* #5000

daiagi:

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #5073 
- [ ] Requires deployment <>

#### Optional

- [ ] I've tested it at </bsx/collection>
- [x] I've tested PR on mobile
- [ ] I've found edge cases

#### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=EfmnRhHaQqfT3phm4cUCHCU3gFVDoSPR1U9WXzMRQBMqZ4L)

#### Community participation

- [ ] [Are you at KodaDot Discord?](https://discord.gg/35hzy2dXXh)

## Screenshot 📸

image
image
image

### DIFF
--- a/components/collection/CollectionInfo.vue
+++ b/components/collection/CollectionInfo.vue
@@ -1,40 +1,42 @@
 <template>
-  <div class="is-flex is-justify-content-space-between">
+  <div class="is-flex is-justify-content-space-between flex-direction">
     <div class="is-flex is-flex-direction-column limit-width">
       <div v-if="collectionInfo?.currentOwner" class="is-flex mb-2">
-        <div class="mr-2">Creator:</div>
-        <nuxt-link
-          class="is-size-6 break-word mr-2 has-text-link"
-          :to="`/${chain}/u/${address}`">
-          <span data-cy="identity-display"> {{ shortenedAddress }}</span>
-        </nuxt-link>
+        <div class="mr-2">{{ $i18n.t('activity.creator') }}</div>
+        <IdentityIndex
+          :address="address"
+          hide-identity-popover
+          :show-clipboard="false"
+          class="has-text-link" />
       </div>
       <div>
         {{ collectionInfo?.meta.description }}
       </div>
     </div>
     <div>
-      <div class="columns">
+      <div class="columns is-mobile">
         <div class="column">
-          <CollectionInfoLine title="Network" :value="chain" />
+          <CollectionInfoLine
+            :title="`${$i18n.t('activity.network')}`"
+            :value="chain" />
           <CollectionInfoLine
             title="Items"
             :value="representation(stats.collectionLength)" />
           <CollectionInfoLine
-            title="Owners"
+            :title="`${$i18n.t('series.owners')}`"
             :value="representation(stats.uniqueOwners)" />
         </div>
         <div class="column">
-          <CollectionInfoLine title="Floor">
+          <CollectionInfoLine :title="`${$i18n.t('activity.floor')}`">
             <CommonTokenMoney
               :value="stats.collectionFloorPrice"
               inline
               :round="2" />
           </CollectionInfoLine>
-          <CollectionInfoLine title="Best Offer">
+          <CollectionInfoLine :title="`${$i18n.t('activity.bestOffer')}`">
             <CommonTokenMoney :value="stats.bestOffer" inline :round="2" />
           </CollectionInfoLine>
-          <CollectionInfoLine title="Volume">
+          <CollectionInfoLine :title="`${$i18n.t('activity.volume')}`">
             <CommonTokenMoney
               :value="stats.collectionTradedVolumeNumber"
               inline
@@ -46,12 +48,14 @@
   </div>
 </template>
 <script setup lang="ts">
-// TODO onChain identity
-// TODO translation
-// TODO which network colection belongs to?
-import shortAddress from '@/utils/shortAddress'
+// TODO
+// which network collection belongs to?
+// offers on RMRK? what to show instead?
+// limit height of collection description?
+
 import CollectionInfoLine from './collectionInfoLine.vue'
 import CommonTokenMoney from '@/components/shared/CommonTokenMoney.vue'
+import IdentityIndex from '@/components/identity/IdentityIndex.vue'
 import {
   useCollectionDetails,
   useCollectionMinimal,
@@ -67,7 +71,6 @@ const chain = computed(
       .text
 )
 const address = computed(() => `${collectionInfo.value?.currentOwner}`)
-const shortenedAddress = computed(() => shortAddress(address.value))
 
 const { collection: collectionInfo } = useCollectionMinimal({
   collectionId: collectionId.value,
@@ -90,5 +93,33 @@ const representation = (value: number | undefined): string => {
 @import '@/styles/abstracts/variables';
 .limit-width {
   max-width: 320px;
+}
+
+.flex-direction {
+  flex-direction: row;
+}
+
+@include mobile {
+  .flex-direction {
+    flex-direction: column;
+    gap: 1rem;
+  }
+
+  .limit-width {
+    max-width: 100%;
+  }
+}
+.has-text-link {
+  a {
+    color: inherit;
+  }
+}
+</style>
+
+<style lang="scss">
+.has-text-link {
+  a {
+    color: inherit !important;
+  }
 }
 </style>
--- a/components/collection/CollectionInfo.vue
+++ b/components/collection/CollectionInfo.vue
@@ -1,0 +1,94 @@
+<template>
+  <div class="is-flex is-justify-content-space-between">
+    <div class="is-flex is-flex-direction-column limit-width">
+      <div v-if="collectionInfo?.currentOwner" class="is-flex mb-2">
+        <div class="mr-2">Creator:</div>
+        <nuxt-link
+          class="is-size-6 break-word mr-2 has-text-link"
+          :to="`/${chain}/u/${address}`">
+          <span data-cy="identity-display"> {{ shortenedAddress }}</span>
+        </nuxt-link>
+      </div>
+      <div>
+        {{ collectionInfo?.meta.description }}
+      </div>
+    </div>
+    <div>
+      <div class="columns">
+        <div class="column">
+          <CollectionInfoLine title="Network" :value="chain" />
+          <CollectionInfoLine
+            title="Items"
+            :value="representation(stats.collectionLength)" />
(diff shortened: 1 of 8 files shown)


exezbcz:
hey hey, nice job 
- I think we can stretch the description even more, so it won´t take too much space vertically 
image
- Afaik there is also a limit so it wont get any longer - solution would also be too put "see more" collapsible option so the screen is not only about the description - don´t implement this, I am just thinking out loud :D 

just to be sure, the social icons are going to be over the banner

daiagi:
> I think we can stretch the description even more, so it won´t take too much space vertically

agree, will do

> just to be sure, the social icons are going to be over the banner

yep, i just waited for the banner to be merged

a couple of follow up questions:

1. RMRK doesn't have offers (does it?) what should we place there instead?
2. @kodadot/internal-dev  where can i get the info of which network the collection belongs to?
3. is "best offer" should be the best offer for the entire collection or for an individual NFT in the collection?


thanks

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
...
+      </div>
+      <div>
+        {{ collectionInfo?.meta.description }}
+      </div>
+    </div>
+    <div>
+      <div class="columns is-mobile">
+        <div class="column">
+          <CollectionInfoLine
+            :title="`${$i18n.t('activity.network')}`"
```
Jarsen136, line 20:
```suggestion
            :title="$t('activity.network')"
```

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import '@/styles/abstracts/variables';
+.limit-width {
+  max-width: 320px;
+}
+
+.flex-direction {
+  flex-direction: row;
```
Jarsen136, line 99: `is-flex` pls use more bulma class
https://bulma.io/documentation/helpers/flexbox-helpers/#docsNav

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+<script setup lang="ts">
+// TODO
+// which network collection belongs to?
+// offers on RMRK? what to show instead?
+// limit height of collection description?
```
Jarsen136, line 54: Will there be a follow-up issue for them?

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
...
+const route = useRoute()
+const { availableChains } = useChain()
+
+const collectionId = computed(() => route.params.id)
+const chain = computed(
+  () =>
+    availableChains.value.find((chain) => chain.value === route.params.prefix)
+      .text
+)
+const address = computed(() => `${collectionInfo.value?.currentOwner}`)
```
Jarsen136, line 73:
```suggestion
const address = computed(() => collectionInfo.value?.currentOwner)
```

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <div class="is-flex is-justify-content-space-between flex-direction">
```
Jarsen136, line 2:
```suggestion
  <div class="is-flex is-justify-content-space-between">
```

row-direction is a default option

Jarsen136 (changes requested):

exezbcz:
@daiagi 
thanks! 
nope, rmrk does not have offers - we can remove the line imo

best offer in this case works like floor price - its a best offer across all nfts - so if one nft has offer and the other dont, the best offer is the one. Its more general approach because we dont have collection offers yet

components/collection/CollectionInfo.vue
daiagi: that alone give TS error

:title="`${$t('activity.network')}`" 
to get rid of the error

components/collection/CollectionInfo.vue
daiagi: I've change the class to mobile-flex-direction-column
i still need it in order to change flex-direction in mobile

daiagi:
> nope, rmrk does not have offers - we can remove the line imo

Done

components/collection/CollectionInfo.vue
daiagi: flex-direction row was indeed redundant
but i still need a class other then bulma provides in order to change flex-direction to column on mobile devices

roiLeo (commented):
from what I can see on design, first [row](https://bulma.io/documentation/columns/sizes/):
- creator name & collection desription should be in a sized 6 column
- collection info line should be in a sized 3 column align to right (offset-3)

note: preview failed?

roiLeo:
> got it, ill fix edit: the grid system gave me troubles, I've implemented instead using max-width

Oh no... what happened? have you tried to wrap in `columns`?

> also @roiLeo , do you have input for me as to how to determine which chain a collection belongs to?

Why would you need that for? can't you check with `prefix`?

daiagi:
> Oh no... what happened? have you tried to wrap in `columns`?

yes, i did, the content overflows the grid and breaking the layout
I think it's because I've set a min-padding between field name and value (such as Network:       Kusama) 
which is mandated by the design
It looks good to me as is right now


> > also @roiLeo , do you have input for me as to how to determine which chain a collection belongs to?
> 
> Why would you need that for? can't you check with `prefix`?

 a user can change the network using the network dropdown, which will change the text in the collection description, but surely the collection doesn't belong to all networks, dies it?

image
image

components/collection/CollectionInfo.vue
Jarsen136: - We will solve this ts error in another issue as it's the most common usage in the codebase.
- There is still some `${$i18n` usage in your PR, pls change them.

roiLeo:
> yes, i did, the content overflows the grid and breaking the layout 

I can see this is doable, I'll try something or we can refactor it latter

> a user can change the network using the network dropdown, which will change the text in the collection description, but surely the collection doesn't belong to all networks, does it?

We don't allow to change network on collection page, I don't know what you are talking about since I don't see Network selector on design. 
Collection Id is by network.


By the way `no-wrap` class already exist and is available globally

daiagi:
> We don't allow to change network on collection page, I don't know what you are talking about since I don't see Network selector on design. Collection Id is by network.

You're right.
My bad, I was focusing more on the component already on the unfinished page then on the figma design
this however, requires no change to the code


> By the way `no-wrap` class already exist and is available globally

Removed it. Thank you


Mergable?

exezbcz:
hey hey, one change
image
- text to bold please 

image
- and this to normal weight


gj!

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,125 @@
...
+  <div
+    class="is-flex is-justify-content-space-between mobile-flex-direction-column gap">
+    <div class="is-flex is-flex-direction-column is-flex-grow-1 max-width">
+      <div v-if="collectionInfo?.currentOwner" class="is-flex mb-2">
+        <div class="mr-2">{{ $t('activity.creator') }}</div>
+        <IdentityIndex
+          :address="address"
+          hide-identity-popover
+          :show-clipboard="false"
+          class="has-text-link has-text-weight-normal" />
```
roiLeo, line 11:
```suggestion
        <nuxt-link
          :to="`/${urlPrefix}/u/${address}`"
          class="has-text-link">
          <IdentityIndex ref="identity" :address="address" />
        </nuxt-link>
```
Something like this will work? Instead of your weird css fix

daiagi: Thanks!
That works

daiagi:
Mergable?

Jarsen136 (approved):

roiLeo (commented):
> Mergable?

might be related with kodadot/snek#142 (or kodadot/snek#33)

components/collection/utils/useCollectionDetails.ts
```diff
@@ -19,6 +21,7 @@ export type CollectionEntityMinimal = {
   currentOwner: string
   type: string
 }
+const chainsSupportingOffers = ['bsx']
```
vikiival, line 24: please use config map :) `*.config.ts`

components/collection/CollectionInfo.vue
```diff
@@ -0,0 +1,106 @@
...
+const representation = (value: number | undefined): string => {
+  if (value == undefined) {
+    return '0'
+  }
+
+  if (value >= 1000) {
+    return (value / 1000).toFixed(2) + 'K'
+  }
+  return `${value}`
+}
```
vikiival, line 84: What does this app do ?

vikiival (changes requested):
👁️

components/collection/CollectionInfo.vue
daiagi: Visual representation of a number.
For example 5237 =>5.23K

components/collection/utils/useCollectionDetails.ts
daiagi: Could you point me to an example please?

components/collection/CollectionInfo.vue
roiLeo: IMO we should show full numbers

daiagi: I followed the figma design.
Plus what about mobile?

  Why you prefer full number?

vikiival: > Why you prefer full number?

It is unified across the app

daiagi:
> please use config map :) *.config.ts

:heavy_check_mark:  done @vikiival 

> IMO we should show full numbers

 :heavy_check_mark:  done @roiLeo @exezbcz 
image

components/collection/utils/useCollectionDeatils.config.ts
```diff
@@ -0,0 +1,5 @@
+const config = {
```
vikiival, line 1: File name has typo

Please follow pattern as other *.config.ts

vikiival (changes requested):

components/collection/utils/useCollectionDeatils.config.ts
daiagi: fixed

vikiival:
pay 50 usd

Issue #4755, opened 2023-01-18
Replace vuex with pinia/useState


preschian:
"Nuxt no longer provides a Vuex integration. Instead, the official Vue recommendation is to use pinia, which has built-in Nuxt support via a [Nuxt module](https://pinia.vuejs.org/ssr/nuxt.html)." https://nuxt.com/docs/migration/configuration#vuex

- pinia support nuxt bridge https://pinia.vuejs.org/ssr/nuxt.html
- useState composables https://nuxt.com/docs/getting-started/state-management

related code:
- https://github.com/search?q=repo%3Akodadot%2Fnft-gallery+getters.&type=code
- https://github.com/search?q=repo%3Akodadot%2Fnft-gallery+dispatch%28&type=code
- https://github.com/search?q=repo%3Akodadot%2Fnft-gallery+commit%28&type=code

roiLeo:
## Progress

- [x] lang #5057 
- [x] wallet #5057 
- [ ] fiat #5138 
- [ ] exploreFilters #5139 
- [ ] assets
- [ ] chain
- [ ] explorer
- [ ] history
- [ ] identity
- [ ] identityMint
- [ ] pinning
- [ ] preferences
- [ ] setting
- [ ] index?

preschian:
@vikiival @roiLeo since we are moving forward with pinia, do we still need `@kodadot1/vuex-options`?

roiLeo:
> do we still need `@kodadot1/vuex-options`?

🤔 question is will it work with Nuxt3? 
one option would be to rewritte it to pinia

vikiival:
> @vikiival @roiLeo since we are moving forward with pinia, do we still need `@kodadot1/vuex-options`?

Good Q
happy to deprecate,
Move majority to statick

yangwao:
<img src="https://media0.giphy.com/media/KtPjuhm119ggU/giphy.gif"/>

roiLeo:
Last piece is on #6210, don't know what happened if we remove code

PR #5139, opened 2023-03-01
✨ migrate exploreFilters store to pinia

roiLeo:
## PR Type

- [x] Refactoring

## Context

- [x] related #5057 & #4755


stores/exploreFilters.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const useExploreFiltersStore = defineStore('exploreFilters', {
+  state: (): State => ({
+    listed: false,
+    owned: false,
+    min: undefined,
+    max: undefined,
+  }),
+  getters: {
+    getListed: (state) => state.listed,
```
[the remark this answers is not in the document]
roiLeo: > I don't think we need `getListed, getOwned, getMin, getMax`. because we can access it directly. wdyt?

Yes! Should work like this too lemme change that

components/explore/filters/StatusFilter.vue
```diff
@@ -26,8 +26,9 @@
 <script lang="ts" setup>
 import useReplaceUrl from './useReplaceUrl'
 import { NeoCheckbox } from '@kodadot1/brick'
+import { useExploreFiltersStore } from '@/stores/exploreFilters'
 
-const { $store } = useNuxtApp()
+const exploreFiltersStore = useExploreFiltersStore()
```
Jarsen136, line 31:
```suggestion
const { listed, setListed, owned, setOwned } = useExploreFiltersStore()
```
I wonder if it could work as well.

Jarsen136 (commented):
otherwise lgtm

components/explore/filters/StatusFilter.vue
roiLeo: ts error `Cannot redeclare block-scoped variable 'listed'`, same for owned

roiLeo: oh nvm it's because we declare `const listed = ...` in this file

roiLeo: I'm not sure it's a good pratice to abstract main store (from multiple exemple I can see), I'd say keep it that way for now.

Otherwise people tend to do something like this for actions:
```js
const { setListed, setOwned } = exploreFiltersStore
```

Jarsen136: Oh, I found a demo here https://pinia.vuejs.org/core-concepts/. That seems a best practice.

image

Jarsen136 (approved):

components/explore/filters/StatusFilter.vue
Jarsen136: We could keep the code that way for now.

PR #5348, opened 2023-03-24
#5167 breadcrumbs for min and max price

### LINKED ISSUE #5167: Price in breadcrumbs and sidebar
so here are few changes, currently we are not showing active price filter as a tag in breadcrumbs 

## Changes 
- when you input price filter, either min or max, or both - there will be breadcrumb displayed 
- the numbers will stay inputted when you click apply - so not resetting, only making the button inactive - this will make more sense on mobile as well 
- apply button will be active once you change the filter or delete one of the parameters. (the price has to make sense, for example min price 7 and max 6 is not possible etc. but I think this is already working fine. )

example: 
- I applied filter min price 1 
image

- the number will stay in the input field 
- button will be now inactive
- breadcrumb is showed 

### add currency 
- it would be also nice to show in what currency is the price denominated 
image
- something like this

### Breadcrumb styling 
- the design is the same, we can collapse the price into one breadcrumb - it will say something like: min 15 KSM, max 20 KSM
- we can later separate but I think this is better

**- When you close the breadcrumb, the numbers you entered in the price section will remain there, but the button will become active because no price filter is there.**

There may be cases for which this mechanism is not a good fit - if there are any problems, please let me know :D I may be wrong.

prachi00:

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #5167
- [ ] Requires deployment <>

#### Optional

- [ ] I've tested it at </bsx/collection>
- [ ] I've found edge cases

#### Had issue bounty label?

- [x] Fill up your KSM address: 
[Payout](https://beta.kodadot.xyz/transfer/?target=EzGc4s9PgCPx1YnF3fqzhLzVHpHMTL4LWPScwpDrR8JKgSU)

#### Community participation

## Screenshot 📸

Screen Shot 2023-03-23 at 9 30 14 PM


prachi00:
Known issue:
Screen Shot 2023-03-23 at 9 30 22 PM
with fluid layout, the content cuts off when value is 2 digits or 3 digits 

@exezbcz Also, I did not disable the apply button as there were issues with keeping track of the previous value as well and when to make it enable, are you okay with how it works right now?

components/shared/BreadcrumbsFilter.vue
```diff
@@ -4,11 +4,31 @@
...
+        class="control c"
+        @close="removeBread('max')">
+        {{ `${$t('Max')}:` }}
+        <CommonTokenMoney :value="value" />
+      </NeoTag>
+      <NeoTag
+        v-else
+        :key="key"
+        class="control d"
+        @close="closeTag(String(key))">
```
vikiival, line 31: Why can't this fit into?
```vue
<NeoTag v-else :key="key" class="control" @close="closeTag(String(key))">
```

exezbcz:
@prachi00 

> the design is the same, we can collapse the price into one breadcrumb - it will say something like: min 15 KSM, max 20 KSM
- can you please make it like that

it wont fit more digits? 
there could be less padding 
image

Yeah, best solution would be to put it on second row and make the apply button smaller  
- layout similar to this
image



soo no disabled button when filter is applied? 
image
- that is pretty nice feature because otherwise you could think that you can click apply again

components/shared/BreadcrumbsFilter.vue
roiLeo: maybe to add label?

prachi00: yeah label is different for them, I'll see if it can be merged in some way

prachi00:
> 

@exezbcz do you mean instead of 2 breadcrumbs, there should be 1 for min and max? 
I think it should be two because you can remove one of them and just have stuff for max to show up, for example right now, you can do max 20ksm without giving any value to min, and still click on apply, so in this case only max breadcrumb would show up

vikiival (approved):
Code looks oki 

Waiting for @exezbcz

exezbcz:
@prachi00 you can click apply when you input only one parameter. - there will be either min, or max - when you apply both, there is min: ... max: ....
The thing why grouping the breadcrumbs is better is that you can than close both with one click.

exezbcz:
@vikiival , yeah, it's mostly okay from my point of view. Still, one specification is missing, which is to make the button inactive when you click 'Apply' and the filter is applied. So, I'm not sure if we should move on and make a follow-up issue.

vikiival:
> @vikiival , yeah, it's mostly okay from my point of view. Still, one specification is missing, which is to make the button inactive when you click 'Apply' and the filter is applied. So, I'm not sure if we should move on and make a follow-up issue.

> you can click apply when you input only one parameter.


I think this is imo fine

vikiival:
Nice job @prachi00 ^-^

yangwao:
💵 $30 @ 32.7 USD/KSM ~ 0.917 $KSM
🧗 [EzGc4s9PgCPx1YnF3fqzhLzVHpHMTL4LWPScwpDrR8JKgSU](https://kusama.subscan.io/account/EzGc4s9PgCPx1YnF3fqzhLzVHpHMTL4LWPScwpDrR8JKgSU)
🔗 [0x1f39b4485c7acd962d585ea6eb6f9f0fca49a41e5372175c49af3cb982c5f2d2](https://kusama.subscan.io/extrinsic/0x1f39b4485c7acd962d585ea6eb6f9f0fca49a41e5372175c49af3cb982c5f2d2)

prachi00:
@vikiival we can merge this one

PR #5423, opened 2023-03-29
Don't merge me yet 

yangwao:
Testing if loading on landing takes while too and we don't show RMRK2 in latest list and sales


yangwao:
ah actually we are showing it on deploy previews, I haven't realized well, merging? 
 
image

PR #5668, opened 2023-04-14
🔧 gtag plugin

roiLeo:
yet another test ⬆️ 

I'm pretty sure nuxt3 plugins doesn't work with `nuxt-edge`


yangwao:
would be nice to get some numbers back well

PR #5815, opened 2023-04-26
Update CODEOWNERS

roiLeo:
IMO the change was breaking review process


yangwao:
hmm, ok for now, we are awating for two ppl in that qa-guild, so looking forward adding it back

PR #5872, opened 2023-05-06
add skeleton when query is loading

### LINKED ISSUE #5749: Default loading dialog on explorer and collections
On slower internet we are showing to users that there are none results, even there are and items will load later afterwards. 
On slower internet it can result bouncing off visitors

Can we show different phrasing, when it's trying actually load items/collections on explorer & collection & profile ?

Like `Wait a while, we are indexing over 1M+ items and we are looking for items by your search parameters right now`

and if query fails (do we check that actually?), then 

`Whoops, we couldn't find anything matching your search.`
`Give it another shot with different parameters or check back later.  New awesome NFTs are added all the time, even now`

image

leo-anderson-x:

## PR Type

- [x] Feature

## Context

- [x] Closes #5749 

#### Had issue bounty label?

- [x] Fill up your KSM address: [Payout](https://beta.kodadot.xyz/transfer/?target=DzUbHCk3Fr3XdCPNKo7uCJvtBH7YfgiFbn4Gr3VmCMiFy1C)

## Screenshot 📸

https://user-images.githubusercontent.com/128157824/236622497-d7843283-eadb-4415-8b5d-848a73351c71.mp4

### <samp>🤖 Generated by Copilot at 93ba4fe</samp>

This pull request adds loading states for the collection list and items grid components, using skeleton components to show placeholders while the data is loading. It also adds two new skeleton components, `CollectionCardSkeleton.vue` and `NeoNftCardSkeleton.vue`, to the `@kodadot1/brick` library, and exports them for reuse.

### <samp>🤖 Generated by Copilot at 93ba4fe</samp>

> _Skeletons appear_
> _`b-skeleton` shapes the cards_
> _Winter of loading_


exezbcz:
nice! works for me 🚀 

wdyt about removing hover on skeleton load? Does that make sense or nah?

components/collection/CollectionCardSkeleton.vue
Jarsen136, line 1: This skeleton logic could be placed into the original component.`components/collection/CollectionDetail.vue`
In this way, we do not need to create another similar component.

libs/ui/src/components/NeoNftCard/NeoNftCardSkeleton.vue
Jarsen136, line 1: This skeleton logic could be placed into the original component. `libs/ui/src/components/NeoNftCard/NeoNftCard.vue`

leo-anderson-x:
> removing hover on skeleton load?

@exezbcz updated

components/collection/CollectionList.vue
```diff
@@ -11,20 +11,28 @@
...
       :default-width="{ small: 16 * 15, medium: 16 * 20, large: 16 * 25 }"
       :mobile-variant="false">
+      <template v-if="isLoading">
+        <CollectionCard
+          v-for="n in SKELETON_COUNT"
+          :key="n"
+          :is-loading="true" />
+      </template>
+
       <div
```
Jarsen136, line 27:
```suggestion
        <CollectionCard
          v-if="isLoading"
          v-for="n in SKELETON_COUNT"
          :key="n"
          :is-loading="true" />

      <div v-else
```

roiLeo (commented):
Coule you please use NeoSkeleton component instead of b-skeleton ?

leo-anderson-x:
@roiLeo updated

components/collection/CollectionCard.vue
```diff
@@ -11,10 +13,16 @@
         :name="collection.name || ''"
         :image="image" />
     </nuxt-link>
+
+    <template v-else>
+      <NeoSkeleton :no-margin="true" :rounded="false" height="112px" />
+      <CollectionDetail :is-loading="true" :nfts="[]" name="" />
+    </template>
```
roiLeo, line 20: would it be possible to keep this outside `CollectionCard` component? maybe just loop through skeleton in `CollectiionList` component (or create `NeoCollectionCardSkeleton`)

components/collection/CollectionDetail.vue
```diff
@@ -32,16 +41,28 @@
...
       </div>
     </div>
+    <div v-else class="is-flex is-justify-content-space-around is-vcentered">
+      <div
+        v-for="n in DESC_SKELETON_COUNT"
+        :key="n"
+        class="is-flex is-align-items-center detail-item column px-5">
+        <NeoSkeleton :no-margin="true" size="medium" position="centered" />
+      </div>
+    </div>
```
roiLeo, line 51: same as precedent comment, I would like to keep this outside component

libs/ui/src/components/NeoNftCard/NeoNftCard.vue
```diff
@@ -94,4 +108,12 @@ withDefaults(
...
 <style lang="scss" scoped>
 @import './NeoNftCard.scss';
+
+.is-absolute {
+  position: absolute;
+}
+
+.top-0 {
+  top: 0;
+}
```
roiLeo, line 118: This doesn't look like it's used somewhere

roiLeo (commented):
Doesn't look perfect on large screen, maybe increase numbers: 12 for collections and 14 for items
Capture d’écran 2023-05-09 à 4 10 36 PM

Capture d’écran 2023-05-09 à 4 18 53 PM

FYI: `no-margin` =  `no-margin="true"`

components/collection/CollectionCard.vue
Jarsen136: If we create another component for the skeleton, the CSS for this card component could not be reused.  What's more, if we want to make some changes to one component, we should also take care of the other related component.

leo-anderson-x:
> Doesn't look perfect on large screen, maybe increase numbers: 12 for collections and 14 for items

@roiLeo updated to: if screen width < 1024 (at this point 4 is enough), show 4, else show 12 collections and 14 items.

components/collection/CollectionCard.vue
leo-anderson-x: Putting it in one file seems to make it easier to synchronize updates to the skeleton when the html structure changes?

roiLeo (commented):
I find it more readable to put skeleton near `<DynamicGrid />` component, so you'll have all 3 options in the same file (data, loading skeleton & empty data)
Did you try to put it on that level?

leo-anderson-x:
> I find it more readable to put skeleton near `<DynamicGrid />` component, so you'll have all 3 options in the same file (data, loading skeleton & empty data) Did you try to put it on that level?

@roiLeo u means the following three options in parallel, right?
```html
<DynamicGrid>...</DynamicGrid>

<template v-if="isLoading">
  <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
</template>

<EmptyResult v-if="total === 0 && !isLoading" />
```
But the skeletons also needs to be wrapped in a DynamicGrid to get the same size as a normal Card (it is dynamically derived from the screen size). So if we place three options in parallel, the result will look like this:
```html
<DynamicGrid>...</DynamicGrid>

<DynamicGrid v-if="isLoading" {...same props as above}>
  <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
</DynamicGrid>

<template v-if="isLoading">
  <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
</template>

<EmptyResult v-if="total === 0 && !isLoading" />
```
This seemed a bit repetitive, so I chose to put the skeletons inside the first `<DynamicGrid />`, rather than alongside the first `<DynamicGrid />` like the `<EmptyResult />`.

roiLeo:
> > I find it more readable to put skeleton near `<DynamicGrid />` component, so you'll have all 3 options in the same file (data, loading skeleton & empty data) Did you try to put it on that level?
> 
> @roiLeo u means the following three options in parallel, right?
> 
> ```
> <DynamicGrid>...</DynamicGrid>
> 
> <template v-if="isLoading">
>   <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
> </template>
> 
> <EmptyResult v-if="total === 0 && !isLoading" />
> ```

Yes I find something like that easier to maintain

> But the skeletons also needs to be wrapped in a DynamicGrid to get the same size as a normal Card (it is dynamically derived from the screen size). So if we place three options in parallel, the result will look like this:
> 
> ```
> <DynamicGrid>...</DynamicGrid>
> 
> <DynamicGrid v-if="isLoading" {...same props as above}>
>   <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
> </DynamicGrid>
> 
> <template v-if="isLoading">
>   <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
> </template>
> 
> <EmptyResult v-if="total === 0 && !isLoading" />
> ```

Hmmmm... not sure I'm following you're idea, more like:
```ts
<DynamicGrid v-if="!isLoading && total">...</DynamicGrid> 

<template v-if="isLoading">
  <!-- use same limit as query -->
  <div v-for="index in limit" :key="index">
      (Here you manage to recreate simple card using same class with skeleton components)
  </div>
</template>
 
<EmptyResult v-if="total === 0 && !isLoading" />
```

components/collection/CollectionList.vue
```diff
@@ -11,20 +11,33 @@
...
     </DynamicGrid>
-    <EmptyResult v-if="total === 0" />
+
+    <DynamicGrid
+      v-else-if="isLoading"
+      :id="scrollContainerId"
+      grid-size="medium"
+      :default-width="GRID_DEFAULT_WIDTH"
+      :mobile-variant="false">
+      <CollectionCard v-for="n in SKELETON_COUNT" :key="n" is-loading />
```
leo-anderson-x, line 36: @roiLeo updated here to reflect my idea:

I need to wrap the CollectionCard Skeleton in a `<DynamicGrid />` with duplicate props because otherwise the Skeleton Card would not have the right width, which is dynamically calculated by the `<DynamicGrid />` based on the container size and set for the child components.

If the duplicated `<DynamicGrid />` is not an issue, then I have no problem with this approach.

roiLeo: With your solution you still kept the logic in `CollectionCard` component (Skeleton & isLoading) and it's working well.
I just feel it's better to have two "different" component but I might be wrong.

components/collection/CollectionList.vue
```diff
@@ -43,6 +57,14 @@ const { $apollo } = useNuxtApp()
 const { urlPrefix, client } = usePrefix()
 const preferencesStore = usePreferencesStore()
 
+const breakPointWidth = 1024
+const isSmallScreen = computed(
+  () => useWindowSize().width.value < breakPointWidth
+)
+const SKELETON_COUNT = isSmallScreen.value ? 4 : 12
```
roiLeo, line 64: or like I said you can use `first` so It would be same size as fetched data

roiLeo: and with that I don't think you'll need breakpoint

libs/ui/src/components/NeoNftCard/NeoNftCard.vue
```diff
@@ -63,6 +64,18 @@
...
+      <div class="media-object nft-media">
+        <div class="is-square image">
+          <NeoSkeleton :rounded="false" full-size no-margin />
+        </div>
+      </div>
+      <div :class="`nft-media-info__${variant}`">
+        <NeoSkeleton size="medium" no-margin />
+        <NeoSkeleton size="small" no-margin width="100px" />
+      </div>
+    </template>
```
roiLeo, line 78: This thing can be moved in `DynamicGrid`?

roiLeo (approved):
✅ works for me anyway

components/collection/CollectionList.vue
```diff
@@ -43,6 +57,14 @@ const { $apollo } = useNuxtApp()
...
+
+const GRID_DEFAULT_WIDTH = { small: 16 * 15, medium: 16 * 20, large: 16 * 25 }
```
vikiival, line 66: move to constants pls

prury:
will check again when it's ready

libs/ui/src/components/NeoNftCard/NeoNftCard.vue
roiLeo: > But this would mean that I need to include `NeoNftCard.scss` in `<ItemGrid />` (although it seems like an unrelated component). Is this acceptable?

Okay, let's move on

vikiival:
@prury should be ready

vikiival (approved):

Jarsen136 (approved):

prury:
> @prury should be ready

Its showing 40 skeletons for both collections and items for me, if that is intended, works well, had to use a bandwidth limiter to test it.
- tested on android, firefox, chrome. 
- responsiveness looking good

yangwao:
pay 30 usd

leo-anderson-x:
> Its showing 40 skeletons for both collections and items for me, if that is intended, works well

@prury Yes, it's intended. We use same size as fetched data

PR #5931, opened 2023-05-11
Custom Link component for NeoNftCard

vikiival:
- :technologist: neo-nft-card takes custom link component
- :zap: use nuxt-link for neo-nft-card

# TL;DR

I've seen in the PR that @roiLeo asked @praachi00 to replace nuxt-link with a in NeoNftCard

The problem is that it was doing ugly reloads.

The fix is really hacky but serves the purpose 

Commit: 6c05aecc81a14442d9dfe72074afd8fe3f67dd2f


roiLeo (approved):
✅ code lgtm

Thank you for this solution, btw `bindKey` shouldn't have prefix path and stuff (If used in another repo it will assume `/${prefix}/gallery/${id}` path exist)

PR #5959, opened 2023-05-14
fix: gallery item image container is not square on tablet

### LINKED ISSUE #5682: Gallery item preview on tablet
image
- the item preview container should be square 
should look like this:
image

if there is a case where img is not square - it is centered in the square container 

happening on devices from 780 and higher

leo-anderson-x:

## PR Type

- [x] Bugfix

## Context

- [x] Closes #5682 

CleanShot 2023-05-14 at 17 44 33

### <samp>🤖 Generated by Copilot at ce54cd9</samp>

Improved image quality and performance in the gallery by removing the `original` prop from the image components. This affects the `GalleryItem.vue` component and its subcomponents.

### <samp>🤖 Generated by Copilot at ce54cd9</samp>

> _`original` gone_
> _`o-image` and `o-carousel`_
> _faster, sharper spring_


components/gallery/GalleryItem.vue
```diff
@@ -32,8 +32,7 @@
                   <MediaItem
                     :key="resource.src"
                     :src="resource.src"
-                    is-detail
-                    :original="isMobile" />
```
preschian, line 36: how about on mobile?

leo-anderson-x: @preschian For square images: there is no difference before and after the PR.

For [non-square images](https://canary.kodadot.xyz/bsx/gallery/4017994657-1):
CleanShot 2023-05-14 at 21 52 40

@exezbcz Regarding the changes in non-square images on mobile devices, I guess this is what you intended, as it was part of your requirements for tablet. and this page's mobile and tablet designs are actually the same.

But if you want to keep the original display logic for non-square images on mobile devices, please let me know.

exezbcz: Yup, that works. We made a full-screen button for the non-square content, so this is expected. Its a standard of how we display content all over the app, let's keep it the same and consistent. 
Thanks! I will check later.

exezbcz: wdyt? Open to discussion :D

roiLeo: I think we already had this kind of discussion in the past but I can't find the related issue.

(Looks like buttons are a bit off on mobile screenshot)

leo-anderson-x: @exezbcz Another option: we keep the same logic as before for displaying images on mobile and tablet (in their original scale, not square), but remove the annoying extra space as shown below. This is not consistent with the Desktop side, but I can think of an advantage: if the height of the photo is smaller than the width, then it is more space efficient on mobile.

vikiival:
Any news with this ? cc @exezbcz ?

exezbcz:
@vikiival I approve this pr :D

prury:
hey i checked this one yesterday and squared images were still not showing properly, can give it another go today

PR #5998, opened 2023-05-15
➖ remove general settings

roiLeo:
I like to delete stuff

## PR Type

- [x] Refactoring

## Context

- [x] Closes #5983 
- [x] Vue3 component

#### Before submitting pull request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality

## Screenshot 📸

Capture d’écran 2023-05-15 à 4 15 50 PM

### <samp>🤖 Generated by Copilot at 4ca5004</samp>

Simplified the settings page by removing unused components and tabs, and added a cache-clearing button to the metadata component for easier debugging.

### <samp>🤖 Generated by Copilot at 4ca5004</samp>

> _The settings page had a makeover_
> _No more tabs for general or `metadata`_
> _But don't worry, you can still discover_
> _A button to clear the cache and reset_
> _It's handy for testing and debugging, you bet_

PR #5999, opened 2023-05-15
➖ remove getSetting

### LINKED ISSUE #5985: Get rid of the `getSettings`
Last piece of removing vuex-options is get rid of the `getSetting` function

Screenshot 2023-05-15 at 13 14 24

roiLeo:
## PR Type

- [x] Refactoring

## Context

- [x] after #5998 
- [x] Closes #5985


Jarsen136 (commented):
There is already a `DEFAULT_PREFIX` for the urlPrefix, so we do need to set `ksm` again

utils/mixins/apiUrlMixin.ts
```diff
@@ -1,17 +1,19 @@
...
+export default class ApiUrlMixin extends mixins(PrefixMixin) {
   get apiUrl() {
-    const endpoint = getChainEndpointByPrefix(
-      this.$store.getters.currentUrlPrefix
-    )
-    return endpoint || this.$store.getters.getSettings['apiUrl']
+    const endpoint: string =
+      getChainEndpointByPrefix(this.urlPrefix) ||
+      getChainEndpointByPrefix('ksm') ||
+      ''
```
Jarsen136, line 16:
```suggestion
const endpoint: string =
      getChainEndpointByPrefix(this.urlPrefix) 
```

utils/WithKeyring.ts
```diff
@@ -44,12 +43,8 @@ export default class WithKeyring extends ChainMixin {
       ss58Format:
         correctFormat(ss58Forever) >= 0
           ? correctFormat(ss58Forever)
-          : correctFormat(this.prefixByStore),
+          : correctFormat('0'),
```
vikiival, line 46:
```suggestion
          : correctFormat('2'),
```

utils/WithKeyring.ts
vikiival, line 1: Do we use this component?

utils/mixins/apiUrlMixin.ts
```diff
@@ -1,17 +1,20 @@
...
 @Component
-export default class ApiUrlMixin extends Vue {
+export default class ApiUrlMixin extends mixins(PrefixMixin) {
   get apiUrl() {
-    const endpoint = getChainEndpointByPrefix(
-      this.$store.getters.currentUrlPrefix
-    )
-    return endpoint || this.$store.getters.getSettings['apiUrl']
+    const endpoint: string =
+      getChainEndpointByPrefix(this.urlPrefix) ||
```
vikiival, line 15: this should be enough, otherwise we it should throw err.
but can be done in different scope

vikiival (commented):
Some clarification

utils/WithKeyring.ts
roiLeo: yes and no, removing it in new commit

utils/mixins/apiUrlMixin.ts
roiLeo: I don't understand your comment but ok

vikiival (approved):

utils/mixins/apiUrlMixin.ts
vikiival: That it should throw error if we cannot find url with prefix

vikiival:
**[heavy breathing]**

PR #5856, opened 2023-05-03
➖ remove b-tooltip component

roiLeo:
One tooltip still need to be migrated but is multilined

## PR Type

- [x] Refactoring

## Context

- [x] Related with #5852 
- [x] Remove `<b-tooltip />` component for now
- [x] Vue3 `Minting` component

#### Before submitting pull request, please make sure:

- [x] My contribution builds **clean without any errors or warnings**
- [x] I've merged recent default branch -- **main** and I've no conflicts
- [x] I've tried to respect high code quality standards
- [x] I've didn't break any original functionality

## Screenshot 📸

Capture d’écran 2023-05-12 à 4 01 50 PM

Screenshot 2023-05-03 at 15-37-28 KodaDot - NFT Market Explorer

### <samp>🤖 Generated by Copilot at 150d681</samp>

Refactored the minting UI and code to use Vue 3 composition API and remove unused components. Deleted `Tooltip.vue` and moved `Minting.vue` logic to `DropUpload.vue`. Simplified the minting settings and improved the code quality.

### <samp>🤖 Generated by Copilot at 150d681</samp>

> _`Minting.vue` is no more, we moved it to `DropUpload.vue`_
> _Refactoring the code with Vue 3, we don't need the Tooltip_
> _Simplifying the settings UI, we improve the user experience_
> _We are the masters of minting, we resist the code complexity_

PR #6013, opened 2023-05-16
fix: offer button issue

### LINKED ISSUE #5971: Offers button
Issue on font-size when button is disabled
Screenshot 2023-05-15 at 08-14-16 flores1 KodaDot - NFT Market Explorer

Offers button isn't visible on mobile/table
Screenshot 2023-05-15 at 08-36-04 Blurred Doggo KodaDot - NFT Market Explorer

leo-anderson-x:

👇 \_\_ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix

## Context

- [x] Closes #5971 

**About "offer button isn't visible on mobile/table" issue:**
Cause: `slide-action` gives height to the parent container so that the offer button (slide entry, full-width-action-button) can gain height to be displayed. In `<GalleryItemOffer />`, however, `slide-action` may lose height for two reasons:
1. `active=false` causes the child component `<NeoTooltip />` to not be rendered
2. Even if `<NeoTooltip />` is rendered, the `position: absolute` set for `.neo-tooltip` in `<GalleryItemActionSlides />` will cause it to lose height

#### Community participation

## Screenshot 📸

CleanShot 2023-05-16 at 20 57 22@2x

CleanShot 2023-05-16 at 20 55 37@2x

### <samp>🤖 Generated by Copilot at b6bf940</samp>

This pull request enhances the UI and UX of the gallery item action slides by applying some CSS styles and fixing a tooltip bug. It affects the files `GalleryItemActionSlides.vue` and `GalleryItemOffer.vue` in the `components/gallery` folder.

### <samp>🤖 Generated by Copilot at b6bf940</samp>

> _To make an offer for a gallery item_
> _You need to see the `NeoTooltip` with some style_
> _But there was a bug that hid it_
> _When the user was inactive or timid_
> _So we changed the condition and added some `o-btn` guile_


components/gallery/GalleryItemAction/GalleryItemActionSlides.vue
```diff
@@ -44,6 +44,11 @@ defineProps<{
       font-size: 1rem !important;
       height: 100% !important;
     }
+
+    .o-btn {
+      font-size: 1rem;
+      height: 100%;
+    }
```
roiLeo, line 51: I don't think it's a good idea to set class from here

roiLeo (commented):
I think issue for button font-size comes from `NeoTooltip` component
We miss these class (taken from beat.kodadot.xyz)
Capture d’écran 2023-05-17 à 9 25 36 AM

leo-anderson-x:
> I think issue for button font-size comes from `NeoTooltip` component

I'm chewing on it.

It seems that the root cause is that we lack a large button with an appropriate text font size, so we often need to override the font size of the large button.

components/gallery/GalleryItemAction/GalleryItemActionSlides.vue
leo-anderson-x: @roiLeo What do you think of the following idea?

Create a separate button component for the Gallery Item Action that inherits from `<NeoButton />` and overrides its font size:
```vue
// <GalleryItemActionButton />

<template>
  <NeoButton class='action-button'>...</NeoButton>
</template>

<style>
  .action-button {
    font-size: 1rem;
  }
</style>
```

Use `<GalleryItemActionButton />` instead of `<NeoButton />` in `<GalleryItemOffer />`. This way, we don't need to look elsewhere for various conditions that override the button style.

The override of the button style in the `NeoTooltip` component was indeed recently removed, which is the reason for this issue, but maybe it's not a good idea to add it back? Because it's hard to understand why the button style is overridden in a Tooltip, and it's hard to predict the impact when trying to modify the overridden style here.

yangwao:
Merged
- https://github.com/kodadot/nft-gallery/pull/5856

leo-anderson-x:
@daiagi 

> yet a third possibility, if you remove the > in the css selector to apply it to all children of slide and not just immediate children

Since we also need to add the `height: 100%` style, applying it to all children would cause other styling issues for the button. If we only apply the font-size to all children, we would still need to target the button separately to set `height: 100%`. Therefore, this approach doesn't differ much from option 2.

> a second possibility is, NeoButton has a is-neo class. you could target that instead of targeting o-btn class

I will skip this option based on @roiLeo 's suggestion.

I chose another way to achieve this. My goal is that if the default style of the button is modified, I can see the modification from the button component itself (perhaps it has an additional class name), rather than having to look for possible overrides in other components. I followed the `full-width-action-button` and created another new class name `action-button`.

roiLeo (changes requested):
Issue is still present on other buttons (LIST, TRANSFER)

Screenshot 2023-05-21 at 16-17-15 Carlos KodaDot - NFT Market Explorer

IMO root issue is comming from `NeoTooltip` disabled view

leo-anderson-x:
> Issue is still present on other buttons (LIST, TRANSFER)

@roiLeo fixed

roiLeo (approved):
Nice! 
✅ code lgtm

Jarsen136 (approved):

Issue #6163, opened 2023-06-05
Continue unlockable items page


yangwao:
Tasks
---
- [x] https://github.com/kodadot/nft-gallery/issues/6204
- [ ] refactor unlockable items by comments in https://github.com/kodadot/nft-gallery/pull/6137#discussion_r1217605814
- [ ] add strip to running collection per drop (reuse same item from galery item/landing)
- [x] #6185 
- [x] #6205 
- [x] #6206 

happy line
---
- [x] #6186 
- [x] #6164
- [x] https://github.com/kodadot/nft-gallery/issues/6171

working version
---
- https://canary.kodadot.xyz/stmn/unlockable

_Originally posted by @daiagi in https://github.com/kodadot/nft-gallery/pull/6137#pullrequestreview-1462245688_

yangwao:
@exezbcz can drop success notification
@daiagi if you want to help with mobile! :)

yangwao:
hmmm seems we are missing phases on unlockable page @Jarsen136 👀

<img width="631" alt="image" src="https://github.com/kodadot/nft-gallery/assets/5887929/5afe73d7-49e2-47eb-9b53-ccf57f0fe3d7">

Jarsen136:
> hmmm seems we are missing phases on unlockable page @Jarsen136 👀
> 
> 
> 
> <img width="631" alt="image" src="https://github.com/kodadot/nft-gallery/assets/5887929/5afe73d7-49e2-47eb-9b53-ccf57f0fe3d7">
> 
> 

These relevant code were commented by @vikiival. I will bring them back soon later

yangwao:
I think it's not anymore p1, yet would be good to recheck

roiLeo:
Done?

vikiival:
> Done?

Cc @yangwao I think its done.

PR #6174, opened 2023-06-06
fix: bring back schedule

Jarsen136:
related with https://github.com/kodadot/nft-gallery/issues/6163#issuecomment-1578566761

image


daiagi (approved):

vikiival:
Sir but that component does not work right?

Jarsen136:
> Sir but that component does not work right?

There is no api for this schedule for now, so we could change the config manually to make it display correct date and phase.

vikiival:
There wont be any yet. 
Every phase starts fron 9am ends 11pm and lasts for hours/mints are gone

Issue #6186, opened 2023-06-08
unlockable touchups 


exezbcz:
Few things that need to be tweaked, not the highest priority - there are also changes in copy.

### button 

<img width="295" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/99e14064-e0bf-45e8-bcc6-156bed7a6fa6">

- make the button as in the new wallet + account sidebar 
- small secondary button 

<img width="494" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/043b19e9-0f4c-40b3-b16d-16c116cee78d">

- align it to the right so the button ends where the divider is

<img width="406" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/8f963fff-7f4a-4358-b397-adcb2979b650">

### copy 
<img width="279" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/71ccecab-5d2b-4580-8642-d6516320dae5">

- not sure what this is displaying - it says 16 minutes - to what? 
- looks like its not mint phase either

<img width="499" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/d46df7c3-70b8-4623-bbc4-e8e257da70f6">
- please change to the same timer as is in the bar + add "left" 
<img width="209" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/3de64f4d-bc57-4737-8e85-f91d379b93e4">

### schedule
- so, if I understand correctly, we won't have 10 minutes window each hour and it will be basically continuous 1-hour windows
- rephrase please the explainer of the schedule that there are only 10 items in every phase - every phase has 60 minutes

### Selected img 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/535c0398-4cbe-4ee3-a657-aad3584bb995)

- under the main image are also previews displayed - option to directly click on them - selected bottom image is shown as the main 
- around the selected image is a 3px blue stroke 

![image](https://github.com/kodadot/nft-gallery/assets/90852205/9a6c0644-ccd6-4d61-b1b6-a7a97f906002)

- hover effect same as on nft-cards (opacity) 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/2b063e8e-997d-443b-9e8e-446ca53b4dc8)

ref #6163

kodabot:
[...]

Issue #5852, opened 2023-05-03
Goodbye Buefy


roiLeo:
## Rewrite [Buefy](https://buefy.org/documentation) components in `libs/ui` using [Oruga-ui](https://oruga.io/documentation/)
part of 
- #2690 

#### In Progress
- [x] replace buefy snackbar & toast #6744 

#### Done
- [x] `<b-progress />` => component or check Bulma #5895
- [x] `<b-carousel />` => `<NeoCarousel />` (Oruga [Carousel](https://oruga.io/components/Carousel.html))  #5884
- [x] `<b-carousel-item />` => `<NeoCarouselItem />` #5884
- [x] `<b-icon />` => `<NeoIcon />` #5858
- [x] `<b-modal />` => `<NeoModal />` #5883 
- [x] `<b-datepicker />` => `<NeoDatepicker />` (Oruga [Datepicker](https://oruga.io/components/Datepicker.html)) #5946
- [x] `<b-checkbox />` => `<NeoCheckbox />` #6000 
- [x] `<b-steps />` => `<NeoSteps />` (Oruga [Steps](https://oruga.io/components/Steps.html)) #5917
- [x] `<b-step-item />` => `<NeoStepItem />` #5917
- [x] `<b-switch />` => `<NeoSwitch />` (Oruga [Switch](https://oruga.io/components/Switch.html)) #5945
- [x] `<b-radio-button />` => `<NeoRadioButton />` (Oruga [Radio](https://oruga.io/components/radio.html)) #5976 
- [x] `<b-autocomplete />` => `<NeoAutocomplete />` (Oruga [Autocomplete](https://oruga.io/components/autocomplete.html)) #6018 
- [x] `<b-loading />` => `<NeoLoading />` (Oruga [Loading](https://oruga.io/components/Loading.html)) #5933
- [x] `<b-tooltip />` => #5856 
- [x] `<b-message />` => notification or Bulma [message](https://bulma.io/documentation/components/message/) - 3 results #6068
- [x] `<b-image />` => `<img />` #5854
- [x] `<b-field />` => `<NeoField />` #5870 
- [x] `<b-input />` => `<NeoInput />` #5870
- [x] `<b-tag />` => `<NeoTag />` check Bulma [Tag](https://bulma.io/documentation/elements/tag/) - 3 results #6155 
- [x] `<b-table />` => `<NeoTable />` (Oruga [Table](https://oruga.io/components/Table.html)) - 8 results (related with #6110, #6144)
[...]

PR #6193, opened 2023-06-08
fix: making mint button active when not connected

### LINKED ISSUE #6185: making mint button active when not connected
ref #6163 

sry @vikiival :D 

- should work the same as the buy button in the gallery item - being in the active state even when the wallet is not connected - after being touched, it should open the wallet connect dialog. 

- situations when the button is inactive:
   - minted out drop
   - reached nft limit

image

Jarsen136:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Needs Design check

- @exezbcz please review

## Context

- [x] Closes #6185

## Screenshot 📸

- [ ] My fix has changed UI

### <samp>🤖 Generated by Copilot at 2ebc940</samp>

Added a modal to connect wallet before minting unlockable NFTs. Refactored `NeoButton` component to support modal functionality.

### <samp>🤖 Generated by Copilot at 2ebc940</samp>

> _To mint an NFT with a lock_
> _You need to connect to your block_
> _So a modal was made_
> _With a `NeoButton` to aid_
> _The user in choosing their stock_


components/collection/unlockable/UnlockableContainer.vue
```diff
@@ -170,11 +171,16 @@ import { timeAgo } from '@/components/collection/utils/timeAgo'
...
 import { endOfHour, startOfHour } from 'date-fns'
+import type Vue from 'vue'
+import { ConnectWalletModalConfig } from '@/components/common/ConnectWallet/useConnectWallet'
+
 const { toast } = useToast()
 
 const Loader = defineAsyncComponent(
   () => import('@/components/shared/Loader.vue')
 )
+const { $buefy } = useNuxtApp()
```
roiLeo, line 182: I would like to see something else, please consider refactoring ref #5852

Jarsen136: I guess buefy modal component should be refactored on issue #5852 . For now, we have to use it to open the wallet modal.

daiagi:
@exezbcz 
1. should the Mint button be disabled when not in the minting window?
image

2. @Jarsen136 @vikiival 
there is no notification that minting didn't happen
3. @Jarsen136 @vikiival 
out of the scope of this PR i guess, but if the logged in account can't mint would it be better/possible to disable the mint button with a tooltip instead of making user wait through the loader for nothing?

Jarsen136:
> @exezbcz
> 
> 1. should the Mint button be disabled when not in the minting window?
>    image

It means the end time of the minting window actually. I will change the misleading text as mentioned at the issue #6186

> 2. @Jarsen136 @vikiival
>    there is no notification that minting didn't happen

There is a notification when the mint failed. I have tried to mint some times, and it works though there is no notification about success information. I have added the notification when minting is successful.

> 3. @Jarsen136 @vikiival
>    out of the scope of this PR i guess, but if the logged in account can't mint would it be better/possible to disable the mint button with a tooltip instead of making user wait through the loader for nothing?
> 
> so generally i think it would be better UX if the button is disabled: a. outside of minting window b. when the connected wallet is not allowed to mint
> 
> wdyt?

It's a good idea to show the reason why the button is disabled. But, the tooltip may not be friendly on mobile I guess.

exezbcz:
> 1. should the Mint button be disabled when not in the minting window?

yes, it should, but since we have 60-minute windows, it's not necessary.



> 3\. out of the scope of this PR i guess, but if the logged in account can't mint would it be better/possible to disable the mint button with a tooltip instead of making user wait through the loader for nothing?

Best case would be to have a tooltip for both cases. 
- when the user Is not allowed to mint, it will show that only one per user
- if the minting window ends, it shows either add to calendar or the disabled button saying minting window is not active
- since most of the user will use this page on mobile, tooltips are useless
- 

> so generally i think it would be better UX if the button is disabled: a. outside of minting window b. when the connected wallet is not allowed to mint

yes

vikiival (commented):
Looks oki

yangwao:
💵 $20 @ 4.29 USD/DOT ~ 4.662 $DOT
🧗 [16SjUbGKSdjCdWTy3NNT3JxbRVGGqD4mwkHpc6BD9U2Rp29Z](https://polkadot.subscan.io/account/16SjUbGKSdjCdWTy3NNT3JxbRVGGqD4mwkHpc6BD9U2Rp29Z)
🔗 [0x1f3031201240d51f04d25ec1678a1e84a36e246c50c391abff9235e0c2750b7f](https://polkadot.subscan.io/extrinsic/0x1f3031201240d51f04d25ec1678a1e84a36e246c50c391abff9235e0c2750b7f)

PR #6319, opened 2023-06-23
chore: introduce no follow links

### LINKED ISSUE #6285: Introduce no-follow on links
We should introduce no-follow on links to 
- items at
  - landing
  - collection
  - explorer
  - profile
- profiles

Ref
---
- #6283

abhisheknaiidu:
fixes #6285

### DIFF
--- a/components/Navbar.vue
+++ b/components/Navbar.vue
@@ -52,7 +52,7 @@
       </div>
     </template>
     <template #end>
-      <nuxt-link to="/stmn/unlockable">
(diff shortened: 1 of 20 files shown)


Jarsen136 (commented):
> We should introduce no-follow on links to 
>- items at
 > - landing
  >- collection
  >- explorer
  >- profile
>- profiles

Most of the nft items link on these page would be opened on current window, so these there is no `nofollow` attribute on tha link tag, as I could see in this PR. It's ok to me.
image


The nofollow missing url I could find is at here. Otherwise lgtm
image

PR #6333, opened 2023-06-26
fix: Robots.txt is invalid

### LINKED ISSUE #6330: Robots.txt is invalid
Screenshot 2023-06-26 at 13 44 28

https://developer.chrome.com/docs/lighthouse/seo/invalid-robots-txt/?utm_source=lighthouse&utm_medium=devtools

first iteration #6286

Jarsen136:
## PR Type

- [x] Bugfix

## Context

- [x] Closes #6330 

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -3,7 +3,6 @@ import * as fs from 'fs'
 import { defineNuxtConfig } from '@nuxt/bridge'
 import SentryWebpackPlugin from '@sentry/webpack-plugin'
 import { manifestIcons } from './utils/config/pwa'
-// import { isProduction } from './utils/chain'
 import { URLS, apolloClientConfig } from './utils/constants'
 
 const baseUrl = process.env.BASE_URL || 'http://localhost:9090'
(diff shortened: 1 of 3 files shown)


Jarsen136:
I just learn that `a disallow line should contain a "path" as opposed to a full URL including the domain` after some search.
That means we could not disallow the page index on these subdomains `beta` `canary` etc only by changing the `robot.txt`.

https://stackoverflow.com/questions/71514881/can-domains-be-used-in-robots-txt

roiLeo (approved):

yangwao:
> That means we could not disallow the page index on these subdomains `beta` `canary` etc only by changing the `robot.txt`.

Or maybe set no-index on those pages based on ENV?

Jarsen136:
> > That means we could not disallow the page index on these subdomains `beta` `canary` etc only by changing the `robot.txt`.
> 
> Or maybe set no-index on those pages based on ENV?

I just added `no-index` meta on all pages except for `kodadot.xyz`. Let's see if it work in some case. 
ref: https://developers.google.com/search/docs/crawling-indexing/block-indexing
image

(As for the pages with prefix `.w.`\ `nft ` \`form`, we still need to add different robot.txt for them, or just leave them there.)

vikiival:
For Workers I made a separated issue 
- https://github.com/kodadot/workers/issues/114

vikiival (approved):
let's give it a shot

PR #6340, opened 2023-06-27
🔧 migrate b-slider to NeoSlider

roiLeo:
part of #5852, tested on `/rmrk/mint`

- `b-slider` => `NeoSlider`
- vue composition api `BasicSlider`
- fixed `AttributeTagInput` error

Screenshot 2023-06-27 at 14-00-25 KodaDot Low fees and low carbon minting KodaDot - NFT Market Explorer


components/rmrk/Create/SimpleMint.vue
```diff
@@ -111,7 +111,7 @@
             data-cy="input-batch-address" />
         </NeoField>
         <BasicSlider
-          v-model="distribution"
+          :model-value="distribution"
```
Jarsen136, line 114: This change would break something and then the variable `distribution` would not be changed any more.

Jarsen136 (changes requested):

components/rmrk/Create/SimpleMint.vue
roiLeo: What solution do you suggest? when using v-model it tells me that I'm missing required prop

Jarsen136: We could try to add props `@change="sliderChangeHandler"` on it, and we could listen to the update event from `BasicSlider ` component.

prury:
took me some time to find the right conditions for the slider to appear 💀

components/rmrk/Create/SimpleMint.vue
roiLeo: removed <BasicSlider />, now it's cleaner and it works

roiLeo:
> took me some time to find the right conditions for the slider to appear 💀 lmk if the distribution itself needs to be tested via minting

I don't even know what "distribution" does

prury:
> > took me some time to find the right conditions for the slider to appear 💀 lmk if the distribution itself needs to be tested via minting
> 
> I don't even know what "distribution" does

haha i think its to supposed to send a specific amount(defined by the slider) of the NFTs you minted to the address list above, but i'm not sure either

Issue #5611, opened 2023-04-12
Update `libs/ui` packages


roiLeo:
- Check out official repo example [nuxt3](https://github.com/histoire-dev/histoire/tree/main/examples/nuxt3) or [vue3](https://github.com/histoire-dev/histoire/tree/main/examples/vue3)
- add `@oruga-ui/oruga-next`
- check components

roiLeo:
closed in #7583

PR #6368, opened 2023-06-30
🔧 libs/ui package

roiLeo:
prepare #5611 

- fix issues from `libs/ui` (`pnpm run story:dev`)
- move `IdentityItem` component


libs/ui/src/components/NeoRadio/NeoRadio.story.vue
```diff
@@ -63,7 +63,7 @@
     </Variant>
 
     <Variant title="NeoRadioButton">
-      <NeoField>
+      <div class="field">
```
vikiival, line 66: Why?

vikiival (approved):

PR #6377, opened 2023-07-03
🔧 new apollo clients

roiLeo:
Idk if it's used


vikiival (approved):
it's natively exposed via static

Issue #4750, opened 2023-01-18
Reduce using nuxt-property-decorator and use composition API


yangwao:
We need to reduce using `nuxt-property-decorator` first and use composition API instead. It will help us to migrate to vue3/nuxt3
https://github.com/nuxt-community/nuxt-property-decorator/issues/83#issuecomment-1245139696. at least for newer components
related: https://github.com/kodadot/nft-gallery/issues/2893

And I think we face this problem also https://github.com/nuxt-community/nuxt-property-decorator/issues/83#issuecomment-1246615441; most of the library is still not ready in nuxt3. And maybe some libs/modules are not compatible also in nuxt-bridge

It seems we have a lot of them (484 mentions)

- https://github.com/kodadot/nft-gallery/search?p=1&q=nuxt-property-decorator

Part of
---
- https://github.com/kodadot/nft-gallery/issues/2690

_Originally posted by @preschian in https://github.com/kodadot/nft-gallery/issues/2690#issuecomment-1280409802_
            
## Ref
- https://github.com/kodadot/nft-gallery/issues/2893

yangwao:
It's improving! 

[465 code results in kodadot/nft-gallery](https://github.com/kodadot/nft-gallery/search?p=1&q=nuxt-property-decorator)

yangwao:
[456 code results](https://github.com/kodadot/nft-gallery/search?p=1&q=nuxt-property-decorator)

<img src="https://media4.giphy.com/media/3osxYhMyxzZCLCJLhe/giphy.gif"/>

yangwao:
[415 code results](https://github.com/kodadot/nft-gallery/search?p=1&q=nuxt-property-decorator) in [kodadot/nft-gallery](https://github.com/kodadot/nft-gallery)

<img src="https://media3.giphy.com/media/yRRh94bjTNdXW/giphy.gif"/>

roiLeo:
Can't manage to use `layout()` with nuxt-bridge

yangwao:
> 178 files

<img src="https://media0.giphy.com/media/3o6ZsVLtrVPHjcxXDa/giphy.gif"/>

yangwao:
👀

![image](https://github.com/kodadot/nft-gallery/assets/5887929/c1f6f7e6-f2d1-4b5a-8e8f-194fd8826ccb)

https://github.com/search?q=repo%3Akodadot%2Fnft-gallery+nuxt-property-decorator&type=code

[...]

PR #6394, opened 2023-07-05
🔧 refact nuxt-property-decorator

roiLeo:
part of #4750 => 123 files

- remove unused components
- migrate some components to vue3 composition api


vikiival (approved):
Looks LGTM

Issue #6392, opened 2023-07-05
migrate/remove unused mixins


roiLeo:
part of 
- #4750

we need to remove old mixins implementation

you can find them in `@/utils/mixins/..` folder

Replace:
- `apiUrlMixin` => `const { apiUrl } = useApi()`
- `prefixMixin` => `const { urlPrefix } = usePrefix()`

roiLeo:
some mixins have been removed in #6405

roiLeo:
- closed by #7117

PR #6458, opened 2023-07-22
migrate/remove inlineMixin  and rmrkVersionMixin 

Jarsen136:

## PR Type

- [x] Refactoring

## Context

- [x] Related #6392 migrate/remove inlineMixin  and rmrkVersionMixin 
- [x] Related #4750 reduce using nuxt-property-decorator


composables/useRmrkVersion.ts
```diff
@@ -3,14 +3,14 @@ import { RemarkVersion } from '@kodadot1/minimark/shared'
...
-  const version = computed<RemarkVersion | null>(() => {
+  const version = computed<RemarkVersion>(() => {
     switch (urlPrefix.value) {
       case 'rmrk':
         return '1.0.0'
       case 'ksm':
         return '2.0.0'
       default:
-        return null
+        return '2.0.0'
```
daiagi, line 13: why?

Jarsen136: I originally wanted to keep the same implementation as on the mixin, but now I think it's not that suitable.

daiagi (approved):
lgtm

preschian (approved):

components/shared/LinkResolver.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <component :is="is">
```
vikiival, line 2: hmm ok

Issue #6474, opened 2023-07-27
Blog followup 


exezbcz:
Few things left after merging 

# Landing page
### articles
![image](https://github.com/kodadot/nft-gallery/assets/90852205/75c85d3a-fe27-4533-8617-5aec418e20b3)
- currently it does not have subarticles, but two main one 
- there should be only one hero article at a time
- other articles should be in sections -  if we don't have enough (at least 3) - we can name it somewhat generally and put it there
- the structure is 1 main, 2, 3, 3...

### Title 
- the title should have gradient underneath 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/8f034766-5d51-48c4-b554-74741e77858f)

### Hero card 
- please lower the line height of the article name. Something like 36px should work 
- the whole card could have a bigger height 
- cut the subheading at around 4 rows 
- make the card fixed, so there is always only one size for this specific screen size 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/612a7172-d17d-4537-9e7e-9f459c8ceecb)
- gradient overflows the card 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/437a44c7-ee65-4f32-b935-e03312519638)

roiLeo:
Design can be found in #6419

roiLeo:
Hey! @exezbcz 
I have a few questions about the design:
- it bothers me that I can't access the article when I try to click on the title, is it made on puposes?
- same question for subarticle, should we make the card clickable (hover & stuff) like we do on landing page (Featured Articles section)
- tag doesn't have `•` on card article, is it normal? maybe it will be done in #6475 
- we don't have show date on card?
- no share button?

lmk your thought

exezbcz:
@roiLeo hey hey, thanks for commenting! 

>it bothers me that I can't access the article when I try to click on the title, is it made on puposes?

- yeah, we can change it. The thing is that the main hero article is only one. The rest are article cards, similar to the ones on the landing page 

>same question for subarticle, should we make the card clickable (hover & stuff) like we do on landing page (Featured Articles section)

- yup, it's in the designs like that, my bad it's not explicitly stated somewhere, it's actually a article component with rounded edges to distinguish that its a blog page 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/e0cf5f1d-ff11-48ce-bce8-46627f458213)



[...]

PR #6505, opened 2023-07-31
🔧 blog page refacto

roiLeo:
could be related with #6474

- fix main title
- fix hero-card tag color
- fix hero-card title line-height
- truncate card text
- fix some sizes

Screenshot 2023-07-31 at 14-57-20 KodaDot - NFT Market Explorer

### DIFF
--- a/pages/blog/index.vue
+++ b/pages/blog/index.vue
@@ -1,11 +1,37 @@
 <template>
   <div class="content">
-    <div class="content-headline">
-      <h1>KodaDot Blog</h1>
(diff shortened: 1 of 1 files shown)


prury:
> @prury if this looks good happy to merge 

Lemme check, just a moment

prury:
it works for this specific blog post, but i do want to see how the other posts will behave using these same CSS settings, lets merge and see? maybe we will need to truncate the title as well later on and mess with the images

PR #6575, opened 2023-08-06
🔧 migrate oruga package to lib/ui

roiLeo:
- In preparation of #5611 
- `@oruga/oruga-ui` is now under `libs/ui` workspace
- fix `NeoIcon` size prop/class

### DIFF
--- a/libs/ui/src/components/NeoIcon/NeoIcon.vue
+++ b/libs/ui/src/components/NeoIcon/NeoIcon.vue
@@ -2,7 +2,7 @@
   <o-icon
     :pack="pack || 'fas'"
     :icon="icon"
-    :size="size || 'small'"
+    :size="size || ''"
     :custom-size="customSize"
     :variant="variant" />
 </template>
@@ -19,6 +19,6 @@ defineProps<{
 }>()
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
 @import './NeoIcon.scss';
 </style>
--- a/libs/ui/src/components/NeoIcon/NeoIcon.vue
+++ b/libs/ui/src/components/NeoIcon/NeoIcon.vue
@@ -2,7 +2,7 @@
   <o-icon
     :pack="pack || 'fas'"
     :icon="icon"
-    :size="size || ''"
+    :size="size || 'default'"
     :custom-size="customSize"
     :variant="variant" />
 </template>
--- a/libs/ui/src/histoire.scss
+++ b/libs/ui/src/histoire.scss
@@ -1,8 +1,6 @@
 /* // @import '@/styles/index.scss'; TODO: how to import this file? */
 @import 'bulma/bulma.sass';
-// @import '@oruga-ui/oruga/dist/oruga.min.css';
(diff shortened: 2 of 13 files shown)


libs/ui/src/components/NeoIcon/NeoIcon.vue
```diff
@@ -2,7 +2,7 @@
   <o-icon
     :pack="pack || 'fas'"
     :icon="icon"
-    :size="size || 'small'"
+    :size="size || ''"
```
Jarsen136, line 5:
```suggestion
    :size="size"
```

libs/ui/src/histoire.scss
```diff
@@ -1,7 +1,9 @@
 /* // @import '@/styles/index.scss'; TODO: how to import this file? */
 @import 'bulma/bulma.sass';
-@import '@oruga-ui/oruga/dist/oruga.min.css';
-@import '@oruga-ui/oruga/dist/oruga-full-vars.min.css';
+// @import '@oruga-ui/oruga/dist/oruga.min.css';
+// @import '@oruga-ui/oruga/dist/oruga-full-vars.min.css';
+@import "~@oruga-ui/oruga/src/scss/oruga";
```
Jarsen136, line 5:
```suggestion
@import "@oruga-ui/oruga/src/scss/oruga";
```

Jarsen136 (approved):
small stuff

otherwise lgtm

libs/ui/src/components/NeoIcon/NeoIcon.vue
roiLeo: default size isn't in `'small' | 'medium' | 'large'`

Jarsen136: > default size isn't in 'small' | 'medium' | 'large'

If so, we should define the correct type for the size props in line 14.  Replacing 'small' with an empty string is not a good solution.

> so would it work if provide empty?
If it does not work, we should provide a proper default value for it.

libs/ui/src/histoire.scss
```diff
@@ -1,7 +1,7 @@
 /* // @import '@/styles/index.scss'; TODO: how to import this file? */
 @import 'bulma/bulma.sass';
-@import '@oruga-ui/oruga/dist/oruga.min.css';
-@import '@oruga-ui/oruga/dist/oruga-full-vars.min.css';
+@import "@oruga-ui/oruga/src/scss/oruga";
```
roiLeo, line 3:
```suggestion
@import '@oruga-ui/oruga/src/scss/oruga';
```

libs/ui/src/components/NeoIcon/NeoIcon.vue
roiLeo, line 5:
```suggestion
    :size="size || 'default'"
```
https://github.com/oruga-ui/oruga/blob/4915c4dbcb1a6c2cd39bc7660d41916b330bc0cd/packages/oruga/src/utils/icons.js#L46

Jarsen136 (approved):

PR #6649, opened 2023-08-10
🔧 vue3 composition 

roiLeo:
- part of #4750 
- `<Layout />`
- `<UserGainHistory />`
- `<Interface />`
- `<PriceRange />`
- `<StatsColumn />`


yangwao:
well cypress gone, I would merge it let's go

image

Issue #6625, opened 2023-08-08
Dance in cart


yangwao:
I would like to see items in cart expanded at most.
Collection name probably smaller size?

At price we can skip those small amounts less than $1 
so instead of 96.5000 KSM -> 96 KSM

Follow design best 

![image](https://github.com/kodadot/nft-gallery/assets/5887929/ab695373-d7d8-4a9f-a712-12d2eb0fc85a)

god-of-js:
👋

Issue #6617, opened 2023-08-08
Cart text truncation 


exezbcz:
![image](https://github.com/kodadot/nft-gallery/assets/90852205/22306bfa-cb99-448e-bacf-d0c76e7c8537)
- the nft name is cut even if there is no trash can
- it would also be nice if the text could be fullsize - taking the container
- so when you hover over the nft it will be cut to the length needed 

![image](https://github.com/kodadot/nft-gallery/assets/90852205/aacaaebc-69c2-44f6-94a2-9440e0de0cc5)

yangwao:
Same ? 
- https://github.com/kodadot/nft-gallery/issues/6625

exezbcz:
@yangwao Hmmm, #6617 is more about the nft name rather than the price. I will look into the collection font size. Not sure whether it would look good. Update soon

kodabot:
ASSIGNED - @god-of-js 🔒 LOCKED -> Friday, August 11th 2023, 13:58:54 UTC -> 24 hours

kodabot:
ASSIGNMENT EXPIRED - @god-of-js has been unassigned.

stephenjason89:
👋

kodabot:
ASSIGNED - @stephenjason89 🔒 LOCKED -> Saturday, August 12th 2023, 15:57:56 UTC -> 24 hours

exezbcz:
@stephenjason89 hello sir, @god-of-js already raised pr for this issue - #6665. There are some changes that need to be made. If he doesn't make them in time, the issue will be closed and you can work on it.

stephenjason89:
hello @exezbcz ohh. I already finished. 
i thought the bot said 
<img width="676" alt="image" src="https://github.com/kodadot/nft-gallery/assets/3810177/48fe35c1-ef3f-4250-8afa-754009b48dea">

kodabot:
ASSIGNMENT EXPIRED - @stephenjason89 has been unassigned.

stephenjason89:
I believe this should be closed @exezbcz

PR #6658, opened 2023-08-10
style(): excess truncation made more flexible to allow larger widths

god-of-js:
Screenshot 2023-08-10 at 3 51 35 PM
Screenshot 2023-08-10 at 3 51 55 PM

- [ ] Removed fixed width from name to allow it to span extra empty space

Fixes: #6625 and #6617 

I didn't round up the number as suggested in #6625 I felt the figures after the dot might be important considering we are working with people's cash.
If that is preferred, I can always add that.


vikiival:
Please follow up official PR teplates, otherwise PR cannot be rewarded

yangwao:
@god-of-js please follow the PR template, closing now. Reopen when you will raise PR w/ PR template

PR #6717, opened 2023-08-15
🔧 Vue3 composition Money component

### LINKED ISSUE #3355: Refactor: Money component
> > It's a shared component. It will work well in the same way on moonsama/rmrk/bsx.
> 
> `!isBsx` 👀

Yeap, I just found out that. May need some extra work.

_Originally posted by @Jarsen136 in https://github.com/kodadot/nft-gallery/issues/3336#issuecomment-1176341491_

Since we're introducing multiple currencies, Money component should be refactored to reflect that in a way which is more robust that current implementation.

roiLeo:
- part of #4750
- close #3355 

<s>⚠️ ⚠️ ⚠️ </s>

### DIFF
--- a/components/shared/format/Money.vue
+++ b/components/shared/format/Money.vue
@@ -14,24 +14,6 @@
 import { checkInvalidBalanceFilter, roundTo } from '@/utils/format/balance'
 import formatBalance from '@/utils/format/balance'
 
-const round = (value: string, limit: number, disableFilter: boolean) => {
-  const number = Number(value.replace(/,/g, ''))
-  if (disableFilter) {
-    return parseFloat(number.toString())
-  }
-  return roundTo(value, limit)
-}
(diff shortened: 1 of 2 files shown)


components/shared/format/Money.vue
```diff
@@ -1,55 +1,40 @@
...
-      if (disableFilter) {
-        return parseFloat(number.toString())
-      }
-      return roundTo(value, limit)
-    },
-  },
+const finalValue = computed(() => {
+  return roundTo(
+    formatBalance(checkInvalidBalanceFilter(props.value), decimals.value, '')
+  )
```
vikiival: I would personally move it below props definition

vikiival (approved):
LGTM otherwise

yangwao (approved):

PR #6730, opened 2023-08-16
fix: initial query running twice causing duplicates

### LINKED ISSUE #6677: Same item twice in explore feed
image
- added two items to cart - number of items 2 
image
- scrolled bit more to see the same selected items 

- why is this happening? Looks like more same items are there.

stephenjason89:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context
Issue was caused by 
image

together with
image

Upon navigating to explore items, the query runs already to fetch page 1 but at the same time, breadCrumbsFilter onMounted event sets `listed=true `

Thus making the watcher to run the query again for page 1 which causes the duplicates.

The fix is simple, if you want to set a default filter which is `listed=true` 
It should be done before the watcher gets executed. 

On a separate issue, BreadCrumbsFilter onMounted should also be refactored and all related components that relies on that.
It may cause unnecessary fetching tricking the page that the user set a listed=true filter

- [x] Closes #6677 

#### Optional

- [ ] I've found edge cases

- [x] Fill up your DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer?target=16UcV9V6nVvPYdHz98ymUKmNLkzjCEU5sbKJMi7hxYyTHjzR&usdamount=100&donation=true)

#### Community participation

## Screenshot 📸

### <samp>🤖 Generated by Copilot at 769abca</samp>

Refactored the code for opening the shopping cart modal from the navbar. Moved the logic and configuration from `ShoppingCartModalConfig.ts` and `ShoppingCartButton.vue` to `Navbar.vue` to improve code organization and performance.

### <samp>🤖 Generated by Copilot at 769abca</samp>

> _`ShoppingCartButton`_
> _Simpler, no more `emit` -_
> _Autumn leaves falling_


stephenjason89:
> Hey,
> 
> I am really happy that you managed to find a bug, it was a really tricky one! (good job)
> 
> However, I do not like the implementation fix.
> 
> What I would personally do is:
> 
> * make separated watcher for listed
> * check if oldVal (watch has args (val, oldVal)) is undefined, if yes do nothing
> * if not resetFilter()

Thank you @vikiival, I agree that the fix can be improved.
I was just thinking of 2 scenarios that this watcher might not work.

1st if by any chance, we wouldn't want any default values set. Then if the user tried to set listed=true then it will fail. Since value is from undefined and watcher only runs when oldValue is not undefined.

 
Second, if next time we want to set more default filters not just listed=true, then this issue will arise again.

This fix only catches the listed=true 
> * make separated watcher for listed
> * check if oldVal (watch has args (val, oldVal)) is undefined, if yes do nothing
> * if not resetFilter()
but ofcourse we can also do the skiping of resetFilter when oldValue = undefined on collection & owned

because after you manually unset the filter the value changes to 
- 'false' instead of undefined for 'listed' and 'owned' 
- an empty string '' for collection
- and an empty array for sort

thus making the condition correct to not run resetFilter if oldValue is undefined.

But for other filters like min, max, search. If by any chance, we want to have it set as a default value next time, then this issue may arise again. 

This gets trickier when you try to set multiple default filters. We'll have to check if all of the default filters set came from an undefined value and if one of them is not then it means the filter was set by a person. Assuming we always set a default filter which may not be the case.


Please advice want you want me to change.

Thank you so much @vikiival

yangwao:
cc @vikiival

components/items/ItemsGrid/ItemsGrid.vue
```diff
@@ -62,6 +62,9 @@ import ItemsGridImage from './ItemsGridImage.vue'
 import { useFetchSearch } from './useItemsGrid'
 
 const isLoading = ref(true)
+const route = useRoute()
+route.query.listed = 'true'
```
Jarsen136, line 66: I'm afraid it's not an ideal solution. What if user visit the url with `?listed=felse`


`https://canary.kodadot.xyz/ksm/explore/items?listed=false`

Jarsen136 (changes requested):

vikiival:
> cc @vikiival 

Cc @yangwao 

https://github.com/kodadot/nft-gallery/pull/6730#issuecomment-1680280431

TL;DR i wont merge it like this. Solution is not good and in short term it could create more bugs.

stephenjason89:
> > cc @vikiival
> 
> Cc @yangwao
> 
> [#6730 (comment)](https://github.com/kodadot/nft-gallery/pull/6730#issuecomment-1680280431)
> 
> TL;DR i wont merge it like this. Solution is not good and in short term it could create more bugs.

@vikiival please check my comment, would love to revise it but need feedback on how you'd want me to implement it.
https://github.com/kodadot/nft-gallery/pull/6730#issuecomment-1680374742

vikiival:
> Current OP is a quick & valid fix.

https://github.com/kodadot/nft-gallery/pull/6730#discussion_r1299184618

vikiival:
> Since value is from undefined and watcher only runs when oldValue is not undefined.

So when you swap from undefined to true, watch callback is never turned on?

stephenjason89:
> > Since value is from undefined and watcher only runs when oldValue is not undefined.
> 
> So when you swap from undefined to true, watch callback is never turned on?
> 

Yes, if we implement this

> - check if oldVal (watch has args (val, oldVal)) is undefined, if yes do nothing

components/items/ItemsGrid/ItemsGrid.vue
stephenjason89: You are correct. I forgot to add an if condition `route.query.listed == undefined`

Anyway, let's discuss a better implementation on this with @vikiival

stephenjason89:
> > Current OP is a quick & valid fix.
> 
> https://github.com/kodadot/nft-gallery/pull/6730#discussion_r1299184618

We can address address this with an `if(route.query.listed == undefined)` but would love to solve this with something much better that would work with everything mentioned here https://github.com/kodadot/nft-gallery/pull/6730#issuecomment-1680374742

Thank you for looking into this issue. Looking forward to hearing from you 😄

stephenjason89:
@vikiival @yangwao , hello.
What I end up doing is adding a setDefaultUrl in useReplaceUrl composable.

This handles setting all the default filters
so we can just add this to the component that should have a default filter
image

You can also add more defaults if needed and it will handle it.
image

Since this is specific for explore/items i also added in the function
image
Though we can remove this, and it's up to the developer who'll use the composable to check what they need.

Only downside is that i'm adding a `default=true` param in the url.
This is to check if we have initially set default filters. Thus skipping double execution of query.

components/items/ItemsGrid/useItemsGrid.ts
```diff
@@ -89,9 +89,12 @@ export function useFetchSearch({
       () => route.query.max,
       () => route.query.owned,
       () => route.query.collections,
+      () => route.query.default,
     ],
-    () => {
-      resetSearch()
+    (val, oldVal) => {
+      if (!(!oldVal[7] && val[7] === 'true')) {
```
roiLeo, line 95: this is smelly

roiLeo (commented):
I wouldn't do that way, maybe check `exploreFiltersStore` default value

Jarsen136 (commented):
I suggest closing this PR until we find a better solution that doesn't rely heavily on hard coding.

components/items/ItemsGrid/useItemsGrid.ts
stephenjason89: yeah hahaha, didn't like it either.

stephenjason89:
> I suggest closing this PR until we find a better solution that doesn't rely heavily on hard coding.

how about we implement the initial fix?
it can be just a 2 liner quick fix for the meantime to avoid double execution of query resulting to duplicates
```
const route=useRoute()
if(route.query.listed == undefined) route.query.listed = 'true'
```

vikiival:
> it can be just a 2 liner quick fix for the meantime to avoid double execution of query resulting to duplicates

Yes, but the javascript leart me not to mutate values directly as it can cause weird behaviour

stephenjason89:
@vikiival I understand.
@roiLeo @Jarsen136

Please help me come up with a possible solution guys.  

Let's discuss further
1. Do we want only listed as a default filter? Or do we want to set other filters as default as well?
2. Do we want it shared across 
  - CollectionList.vue
  - Activity.vue 
  - ItemsGrid.vue
  Like the current implementation? It is set by the BreadCrumbsFilter and used on those components. Or should they have 
  different set of default filters?

vikiival:
> Thank you so much guys for looking in to my PR

Please fix deepscan


image

stephenjason89:
@vikiival Thanks done.
I just don't know how deep scan relates those issue to this PR

I'm confused, PulseChart is completely unrelated.
image

Jarsen136:
> > Please fix deepscan
> 
> @vikiival Thanks done. I just don't know how deep scan relates those issue to this PR
> 
> I'm confused, PulseChart is completely unrelated. 

Update the branch to the lastest could help.

stephenjason89:
> > > Please fix deepscan
> > 
> > 
> > @vikiival Thanks done. I just don't know how deep scan relates those issue to this PR
> > I'm confused, PulseChart is completely unrelated.
> 
> Update the branch to the lastest could help.

Thank you @Jarsen136 😃

stephenjason89:
@Jarsen136 do you want me to revert changes to PulseChart & create separate issue and PR?

But deepscan is right PulseChart was wrong because variables weren't initialized. However, based on the notes, i believe that originally the developer want line 11 outside of the function
image

Jarsen136:
I tested it again, but the bug has still not yet been solved. 
https://deploy-preview-6730--koda-canary.netlify.app/ksm/explore/items

Kapture 2023-08-21 at 21 15 03


TBH, the new solution could not solve the problem. It would also introduce more complexity in the meantime.
image

Jarsen136:
> I suggest closing this PR until we find a better solution that doesn't rely heavily on hard coding.

vikiival:
@stephenjason89 can you please open a new pr?

stephenjason89:
yes @vikiival I will. 
Thanks @Jarsen136 for pointing that out.

PR #7080, opened 2023-08-31
feat: more accessible icons

### LINKED ISSUE #6988: Increase clickable area of the close (cross)
image

- Please increase the container size to 24px without changing the icon size itself - The activation will therefore be greater.

thnaylor:

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #6988

#### Optional

- [ ] I've found edge cases

- [x] Fill up your DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=123PCJXhjb15i6JwVVRGd7KvN2sSNZrtPjr5doJq9oWctoTt)

#### Community participation

- [x] [Are you at KodaDot Ecosystem Telegram?](https://t.me/thnaylor)

## Screenshot 📸

### <samp>🤖 Generated by Copilot at 6649f01</samp>

This pull request enhances the `NeoButton` and `NeoModalHead` components in the `ui` library. It adds a new `noBorder` prop to the `NeoButton` component, which removes the border and the background from the button. It also updates the `NeoModalHead` component to use the `noBorder` and `small` props for the close button, and changes the icon to `x`. These changes improve the appearance and consistency of the modal dialog.

### <samp>🤖 Generated by Copilot at 6649f01</samp>

> _Oh, we're the crew of the Neo ship, and we code with skill and flair_
> _We make our buttons look so slick, with `no-border` here and there_
> _We heave and ho, and on we go, to match the design we see_
> _We change the `xmark` to an `x`, and the `NeoModalHead` padding_


thnaylor:
@exezbcz wdyt?

So basically all that's happening is I'm adding a class to remove the border from the buttons we already have which implement icons from Oruga. I feel is the quickest consistent fix. Another option for consistency would just be to create a custom icon button with a minimum required height and widths.

I think at some point you will need to start replacing all the icons around the site with more accessible ones so best to find a solution we can consistently use on the design system and in the code.

The only real con about this is other custom icons don't work well with the Oruga buttons so the standard Font Awesome ones are being used.

Current: 27xW used.

Examples:

**x-small button icon: 27px height x (width of icon)**
image

**small button: 36px height x (width of icon)**
image

**medium button: +-49px height x (width of icon)**
image

**etc...**

exezbcz:
> I think at some point you will need to start replacing all the icons around the site with more accessible ones so best to find a solution we can consistently use on the design system and in the code.

We use fa icons in the designs
ad size 
image
- The best would be to use 24px containers with icon on average 20px wide and high 

this is IMO too small 
image

In the case of sidebars: 
image
- 20px is the smallest we might use. The icon is then 15x15px 

what do you think? cc @thnaylor

thnaylor:
@exezbcz Okay good!

This probably won't fit into our normal NeoButtons well so let me run some tests later today with an alternative method and then i'll build a new example for you. 👍

exezbcz:
@thnaylor we can make exception for the neobutton, the icon is just addition - its not the click area since its wrapped in button

thnaylor:
@exezbcz 

So this one is the smallest NeoButton with click area around 27px by 27px and a "fa-lg" (15px) icon in it

This is probably the best I'll get it using what we already have in the system.

min clickable area: 27px
min icon height: 15px

Exmaple:
image

Example with multiple other icons:
image

exezbcz:
> This is probably the best I'll get it using what we already have in the system.

from what I understand, you are using font awesome pack right? 
image
- few little visual things - the thing is not centered and its not a proper X-cross, Please replace it by this icon https://fontawesome.com/icons/xmark?f=sharp&s=regular

otherwise nice job, looking forward to see the final result!

thnaylor:
@exezbcz should be good now.

- using FA xmark icon
- centered

image

exezbcz:
@thnaylor yes! 
image
- I believe last part is hover effect and its done

exezbcz:
this is not the hover effect for icons :D 
image

thnaylor:
haha soz I was confused because I actually removed the button hover on this so thought you meant that.

I will do the icon only.

exezbcz:
@thnaylor yes this 
in light mode - black to k-grey
in dark mode - white to k-hovergrey

thnaylor:
@exezbcz 

I can't do 2 different classes on thesame icon.

The way it works is we have 1 class e.g "k-grey" and in dark mode its set to '#6b6b6b' and light mode it's set to '#999999'.

k-hovergrey is set the same value for both light and dark mode '#6b6b6b' in the theme.

exezbcz:
@thnaylor its okey then

thnaylor:
@exezbcz if it helps I had a look now and can make it the same as what we already have. 

Look like it's in the theme as 'link-hover'.

exezbcz:
@thnaylor yup

exezbcz:
hmm weird, the icon does not changes with cursor pointer 

https://github.com/kodadot/nft-gallery/assets/90852205/35ece28a-55a5-47ca-81d8-dbde3f7763ac

thnaylor:
@exezbcz Wow, good catch! 🤣 

building again now

libs/ui/src/components/NeoButton/NeoButton.scss
```diff
@@ -211,6 +218,23 @@
...
+
+.o-btn { 
+  @include ktheme() {
+    &:hover {
+      .o-icon {
+        background: transparent;
+        color: theme('link-hover');
+      }
+    }
+  }
```
roiLeo, line 237: this is smelly

roiLeo (commented):
Why did you replace prop on `NeoButton` with `NeoIcon`?
I would just increase padding if we only need to have a larger area

thnaylor:
@roiLeo When using a size small NeoButton the icon sizes are very small and I could not see a good way to fix that so best other option was to use a NeoIcon in the slot and add css to make it work like the button icon.

- Another option which may be cleaner is to remove this all from the NeoButton and move it to the NeoIcon instead and have a appropriate prop like "touch" which will apply the css to make it function and look like the design specification wants?

roiLeo:
> @roiLeo When using a size small NeoButton the icon sizes are very small and I could not see a good way to fix that so best other option was to use a NeoIcon in the slot and add css to make it work like the normal NeoButton icon.

Since the button is the icon, you can add `size` prop on `NeoButton` and it will change icon size cf [oruga doc](https://oruga.io/components/button.html)

> Another option which may be cleaner is to remove this all from the NeoButton and move it to the NeoIcon instead and have a appropriate prop like "touch" which will apply the css to make it function and look like the design specification wants?

- `NeoButton` for buttons and `NeoIcon` for icons

thnaylor:
@roiLeo I took a look at the button sizes and docs.

Designer wants something no smaller than 24x24px and the icon must remain the same size with same functionality.

These are the 2 options that I can make from what we have. Without variant="text" the button size is too big and with variant="text" it is too small.

Example:

| Button | props | Dimension |
| --------------- | ---- | ------------ |
| image | no-border size="medium" | 45x45 |
| image | variant="text" size="medium" | 15x21 |

The only size button the designer likes is the size="small" button (27x27) but the icon needs to be size="medium".

Example:

image

Do you have any suggestion on how to achieve a small button with medium size icon that would be sufficient?

roiLeo:
> Do you have any suggestion on how to achieve a small button with medium size icon that would be sufficient?

> Please increase the container size to 24px without changing the icon size itself - The activation will therefore be greater.

I would just add `p-2` class on the button, keep it simple.

thnaylor:
@roiLeo Thanks. Maybe I'm just overthinking too much 😭 🤣 

Looks decent:
image

prury:
its position has been shifted a bit to the left but the clicking area is indeed bigger
@thnaylor please sync pr

roiLeo (approved):

prury (approved):

yangwao:
pay 50 usd

yangwao:
💵 $50 @ 4.23 USD/DOT ~ 11.82 $DOT
🧗 [123PCJXhjb15i6JwVVRGd7KvN2sSNZrtPjr5doJq9oWctoTt](https://polkadot.subscan.io/account/123PCJXhjb15i6JwVVRGd7KvN2sSNZrtPjr5doJq9oWctoTt)
🔗 [0x210e58c97113c6fe9ad8d5fae5392e9d95f6aa5fd1175d29ef3172837efaa363](https://polkadot.subscan.io/extrinsic/0x210e58c97113c6fe9ad8d5fae5392e9d95f6aa5fd1175d29ef3172837efaa363)

PR #7251, opened 2023-09-12
refactor: [nuxt3] - fa icons

preschian:
**Thank you for your contribution** to the [KodaDot - One Stop Shop for Polkadot NFTs](https://kodadot.xyz).

👇 __ Let's make a quick check before the contribution.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Related #7034
- [x] set fasr (fa-sharp regular) as default icon instead of fas (fa-classic solid)
- [x] use fass (fa-sharp solid) instead of fas (fa-classic solid)
- [x] since we are using fa-kit to set fa, we don't need to specify some icon first in `plugins/icons.ts`

### <samp>🤖 Generated by Copilot at 7305b9e</samp>

This pull request updates various components to use the custom `fass` icon pack for better visual consistency and clarity. It also removes unnecessary icon pack attributes that are redundant or conflict with the default icon pack. These changes simplify the code and improve the appearance of the icons.

### <samp>🤖 Generated by Copilot at 7305b9e</samp>

> _We're sailing with the `fass` icon pack_
> _It makes our icons sharper and sleek_
> _So pull on the rope and don't look back_
> _We'll update all the components this week_

PR #7254, opened 2023-09-13
chore: adjust css deep selector

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Fixes `[@vue/compiler-sfc] ::v-deep usage as a combinator has been deprecated. Use :deep(<inner-selector>) instead.` for nuxt3

image

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1xjvRADwdJcnmUCLWerEHR4iGCT5EBTm4D4fzLLg4LcAC9p)

### <samp>🤖 Generated by Copilot at 9fd8cf7</samp>

This pull request adds the `eslint-plugin-vue-scoped-css` dependency and its rules to the project to improve the linting and quality of scoped CSS in Vue components. It also updates the syntax and usage of `v-deep` in several files to follow the recommended argument syntax instead of the combinator syntax. The `pnpm-lock.yaml` file was updated accordingly.

### <samp>🤖 Generated by Copilot at 9fd8cf7</samp>

> _Oh we're the coders of the `v-deep` syntax_
> _We follow the rules and we don't relax_
> _We use the argument and not the combinator_
> _Or else the linter will give us an error_


roiLeo (approved):
Hmm.. I would do this change on nuxt branch since the config isn't the same for vue3.
For me, deep selector is a bad practice.

otherwise lgtm

preschian:
> Hmm.. I would do this change on nuxt branch since the config isn't the same for vue3

I think as long as the features/specs still have backward compatibility with nuxt2, it's better to update to our nuxt2 also. If there is an eslint rule for that, it will help us to prevent incoming deprecated rules to nuxt2 PR. And, we don't need to resolve conflicts when updating the nuxt3 branch

I just realized I think I should create this PR again for our nuxt2: https://github.com/kodadot/nft-gallery/pull/7251

We create PR to nuxt3 if the features are not backward compatible with nuxt2. For example, NeoModal or NeoNotification, because that PR is not backward compatibility with our nuxt2

daiagi (approved):

roiLeo:
let's go 🚀

Issue #7225, opened 2023-09-09
Fix GraphQL calls in Nuxt3


yangwao:
We need to review GraphQL calls on pages after migration. 
Could you write down which one you plan to fix? 
Make PR against `nuxt` branch.

top of my head
---
- collection view - https://nuxt.kodadot.xyz/ahk/collection/158
- Gallery item 
- Explorer items view - https://nuxt.kodadot.xyz/ahk/explore/items?listed=true
- create items - fetch existing collections
- sidebar calls for balances 
- profile items 
- profile collections
- profile activity
- search - navbar
- search - items from particular collection

Fix all GraphQL calls.
---
- page by page & components by components
- feel free add some which I've missed

sivareddyuppathi:
👋

kodabot:
ASSIGNED - @sivareddyuppathi 🔒 LOCKED -> Sunday, September 10th 2023, 20:32:44 UTC -> 24 hours

kodabot:
ASSIGNMENT EXPIRED - @sivareddyuppathi has been unassigned.

roiLeo:
#### Update

- [x] collection view
- [x] Gallery item (Missing `CarouselTypeVisited`)
- [x] Explorer items view 
- [x] create items - fetch existing collections (tested on bsx/snek missing rmrk)
- [x] sidebar calls for balances 
- [x] profile items 
- [x] profile collections
- [x] profile activity

#### TODO

- [ ] search - navbar
- [ ] search - items from particular collection
- [ ] Holders
- [ ] Sales
- [ ] Spotlight
- [ ] Identity Name?
- [ ] Mass Mint
- [ ] and more...

roiLeo:
so far so good

PR #7299, opened 2023-09-19
Fix graphql

nick-dev123:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring

## Context

- [X] Related to #7225

- [X] My contribution builds **clean without any errors or warnings**
- [X] I've merged recent default branch -- **main** and I've no conflicts
- [X] I've tried to respect high code quality standards
- [X] I've didn't break any original functionality

#### Optional

- [ ] I've found edge cases

- [X] Fill up your DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?usdamount=20&target=12ZLwasAvpdxDiiMedr11VeKUyQkY3WfneMkswBCAHVvxAqe)

#### Community participation

- [ ] [Are you at KodaDot Ecosystem Telegram?](https://t.me/kodadot_eco)

## Screenshot 📸

### <samp>🤖 Generated by Copilot at 5776ac9</samp>

Refactored several components and composables to use a common `usePrefix` function to get the `clientName` parameter for GraphQL queries based on the URL prefix. This improves code consistency and readability, and makes the `clientName` parameter reactive. Wrapped some `data` properties in `value` properties and added optional chaining to make them reactive and avoid errors.

### <samp>🤖 Generated by Copilot at 5776ac9</samp>

> _Oh we're the coders of the sea, and we love to refactor_
> _We use the `usePrefix` function to make our code much neater_
> _We wrap our `data` in a `value` and we pass the `clientName`_
> _And we heave away, me hearties, on the count of three_

### DIFF
--- a/components/bsx/Offer/MasterOfferTable.vue
+++ b/components/bsx/Offer/MasterOfferTable.vue
@@ -157,9 +157,10 @@ const fetchCreatedOffers = async () => {
     if (!skipUserOffer.value) {
       variables.id = destinationAddress.value || accountId.value
     }
-    const { data } = await useAsyncQuery(offerListUser, variables)
-    if (data?.offers?.length) {
-      createdOffers.value = data.offers
+    const { client } = usePrefix()
+    const { data } = await useAsyncQuery(offerListUser, client.value)
+    if (data.value?.offers?.length) {
+      createdOffers.value = data.value?.offers
       if (!skipUserOffer.value) {
         incomingOffers.value = []
       }
--- a/components/bsx/Offer/StatsOverview.vue
+++ b/components/bsx/Offer/StatsOverview.vue
@@ -66,8 +66,8 @@ watchEffect(() => {
     )
 
     statsResponse.value = data.value
-    offerStats.value = data.value.offerStats
-    keysObject.value = Object.keys(data.value).filter(
+    offerStats.value = data.value?.offerStats
+    keysObject.value = Object.keys(data.value as object).filter(
       (key) => key !== 'offerStats'
     )
   } catch (e) {
--- a/composables/transaction/transactionBurn.ts
+++ b/composables/transaction/transactionBurn.ts
@@ -43,8 +43,10 @@ export function execBurnTx(item: ActionConsume, api, executeTransaction) {
       ''
     )
     const hasOffers = ref(false)
+    const { client } = usePrefix()
     const { data } = useGraphql({
       queryName: 'acceptableOfferListByNftId',
+      clientName: client.value,
       queryPrefix: 'chain-bsx',
       variables: {
         id: item.nftId,
--- a/composables/useIdentity.ts
+++ b/composables/useIdentity.ts
@@ -27,7 +27,7 @@ export default function useIdentity({
   )
 
   const { data, refetch, loading } = useGraphql({
-    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')),
+    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')).value,
     queryName: 'identityById',
     variables: {
       id: id.value,
--- a/components/bsx/Offer/OfferStats.vue
+++ b/components/bsx/Offer/OfferStats.vue
@@ -33,7 +33,7 @@ const { pending } = useLazyAsyncData('data', async () => {
   try {
     const { data } = await useAsyncQuery(offerList, {})
 
-    offers.value = data.offers
+    offers.value = data.value?.offers
   } catch (e) {
     $consola.error(e)
   }
--- a/components/carousel/utils/useCarousel.ts
+++ b/components/carousel/utils/useCarousel.ts
@@ -17,9 +17,12 @@ export const useCarouselUrl = () => {
   }
 }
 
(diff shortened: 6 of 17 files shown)


components/bsx/Offer/MasterOfferTable.vue
```diff
@@ -157,9 +157,10 @@ const fetchCreatedOffers = async () => {
     if (!skipUserOffer.value) {
       variables.id = destinationAddress.value || accountId.value
     }
-    const { data } = await useAsyncQuery(offerListUser, variables)
-    if (data?.offers?.length) {
-      createdOffers.value = data.offers
+    const { client } = usePrefix()
+    const { data } = await useAsyncQuery(offerListUser, client.value)
```
roiLeo, line 161: missing `variables`

roiLeo (changes requested):

components/bsx/Offer/StatsOverview.vue
```diff
@@ -66,8 +66,8 @@ watchEffect(() => {
     )
 
     statsResponse.value = data.value
-    offerStats.value = data.value.offerStats
-    keysObject.value = Object.keys(data.value).filter(
+    offerStats.value = data.value?.offerStats
+    keysObject.value = Object.keys(data.value as object).filter(
```
Jarsen136, line 70: Are you about to use `as Object`?

composables/useIdentity.ts
```diff
@@ -27,7 +27,7 @@ export default function useIdentity({
   )
 
   const { data, refetch, loading } = useGraphql({
-    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')),
+    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')).value,
```
Jarsen136, line 30: Why? The `clientName` could be a computed value so that it would be reactive.

roiLeo: I've already done that in all queries so we'll need to change many lines of code

Jarsen136: It should be a reactive value, or it would introduce some bugs. I could expand more about how it causes bugs if needed.

I see there is a new implementation on `useGraphql`. However, it should keep the original logic that allows `clientName` to be reactive. 
image

roiLeo: idk why but https://github.com/kodadot/nft-gallery/blob/5ed65d90db3af697bf7e900ede2d0b9e29d9bbfa/composables/useGraphql.ts#L12 doesn't work as excepted I had to pass the client value on each gql call otherwise it didn't work (not reactive, it would keep old prefix)
If you can push a proof of work where `useGaphql` doesn't need to pass `clientName` I would be glad to see it

Jarsen136 (commented):
It's unnecessary to add such a pattern because we already have a fallback value for the `clientName ` inside `useGraphql`

check `composables/useGraphql.ts`

composables/transaction/transactionBurn.ts
```diff
@@ -43,8 +43,10 @@ export function execBurnTx(item: ActionConsume, api, executeTransaction) {
       ''
     )
     const hasOffers = ref(false)
+    const { client } = usePrefix()
     const { data } = useGraphql({
       queryName: 'acceptableOfferListByNftId',
+      clientName: client.value,
```
Jarsen136, line 49: -

Jarsen136 (commented):
It's unnecessary to add such a pattern because we already have a fallback value for the clientName  inside useGraphql

check composables/useGraphql.ts

composables/useIdentity.ts
roiLeo: > @roiLeo Could you pls provide a small case that which function or page could produce this bug? I'm going to solve it but I cannot produce it now.

Check out `useCarouselSpotlight` query, I had snek result that point on bsx elements

Jarsen136: I see. I will raise a PR to fix the root issue.

Jarsen136: > idk why but
> 
> https://github.com/kodadot/nft-gallery/blob/5ed65d90db3af697bf7e900ede2d0b9e29d9bbfa/composables/useGraphql.ts#L12
> 
> doesn't work as excepted I had to pass the client value on each gql call otherwise it didn't work (not reactive, it would keep old prefix)
> If you can push a proof of work where `useGaphql` doesn't need to pass `clientName` I would be glad to see it

> Check out useCarouselSpotlight query, I had snek result that point on bsx elements

Here is the proof of work that we do not need to pass `clientName` for each ggl call. #7306


> > doesn't work as excepted I had to pass the client value on each gql call otherwise it didn't work (not reactive, it would keep old prefix)

As far as I know, the old prefix only occurs on the local dev env because router does not redirect properly. I also find that there are lots of warnings related to the router. Idk how to fix it. If you check any of the preview env of nuxt branch, you will find that it works well.

Jarsen136:
Thank you for your contribution. I would close this PR because we do not need to make these changes for now.
ref: https://github.com/kodadot/nft-gallery/pull/7308 https://github.com/kodadot/nft-gallery/pull/7306

PR #7308, opened 2023-09-20
🔧 useGraphql client

roiLeo:


composables/useIdentity.ts
```diff
@@ -27,7 +27,7 @@ export default function useIdentity({
   )
 
   const { data, refetch, loading } = useGraphql({
-    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')),
+    clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid')).value,
```
Jarsen136, line 30: I would revert this line and keep it as original.
```
clientName: computed(() => (isDotAddress.value ? 'pid' : 'kid'))
```

roiLeo: alright will do

Jarsen136: OK. We could fix the type warning later because this line would cause a bug.

PR #7309, opened 2023-09-20
Nuxt3

roiLeo:
ping @Jarsen136

### DIFF
--- a/composables/useGraphql.ts
+++ b/composables/useGraphql.ts
@@ -26,7 +26,7 @@ export const useQueryParams = ({
 export default function ({
   queryPrefix = '',
   queryName,
-  clientName = '',
+  clientName = computed(() => ''),
   variables = {},
   options = {},
   disabled = computed(() => false),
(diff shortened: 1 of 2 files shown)


composables/useGraphql.ts
```diff
@@ -26,7 +26,7 @@ export const useQueryParams = ({
 export default function ({
   queryPrefix = '',
   queryName,
-  clientName = '',
+  clientName = computed(() => ''),
```
Jarsen136, line 29: It could stay the same as it does not make any different.
```suggestion
  clientName = '',
```

Jarsen136 (approved):

composables/useGraphql.ts
roiLeo: this throw warning if not computed

Jarsen136: ok. Let's move on

Jarsen136 (approved):

PR #7386, opened 2023-09-27
Not connected wallet collection selection

### LINKED ISSUE #7382: Not connected wallet collection selection
### for follow-up issue I guess:

I think it would be good for UX to have in the collection drop-down a menu item that appears only if not logged in.
clicking this menu item would open the log-in modal (similar to shopping cart)
also, it would be nice to have 
if logged in but no collection is available on the account => drop-down item to create a new collection (similar to massmint)

@exezbcz  wdyt?

_Originally posted by @daiagi in https://github.com/kodadot/nft-gallery/issues/7208#issuecomment-1729557314_
            

yes, I will create design and letsgo

Jarsen136:

  ## PR Type

  - [x] Feature
  - [x] Refactor: unify the collection selection process (mass mint & single nft mint )

  ## Needs Design check

  - @exezbcz please review

  ## Context

  - [x] Closes #7382

  ## Screenshot 📸

  - [ ] My fix has changed UI

  ## Copilot Summary
  
### <samp>🤖 Generated by Copilot at 866ce39</samp>

Improved the user experience of the create landing page by requiring login before redirecting to different creation paths. Used a custom function and a hook to handle the login and redirection logic in `CreateLanding.vue`.

  
### <samp>🤖 Generated by Copilot at 866ce39</samp>

> _No more links to guide your way_
> _Only divs that make you pay_
> _If you want to create your fate_
> _You must login or face the gate_


components/create/CreateLanding.vue
```diff
@@ -28,7 +28,7 @@
       </div>
 
       <div v-else class="pack">
-        <nuxt-link class="pack-box" to="/create/nft">
+        <div class="pack-box" @click="gotoPathAfterLogin('/create/nft')">
```
roiLeo, line 31: Why don't you keep `nuxt-link` and use `@click.prevent`?

`<div>` with link is bad semantic

daiagi:
Can we add "create collection" in the "select collection" drop down in the nft creation form as well?

Like in massmint

components/create/CreateLanding.vue
```diff
@@ -49,18 +51,29 @@
...
+const instance = getCurrentInstance()
+const { doAfterLogin } = useDoAfterlogin(instance)
 const { urlPrefix } = usePrefix()
 const route = useRoute()
 const router = useRouter()
+
+const gotoPathAfterLogin = (path: RawLocation) => {
+  doAfterLogin({
+    onLoginSuccess: () => {
+      router.push(path)
```
daiagi, line 73: If it remember correctly we have some util "navigate To"
Is it suitable here?

exezbcz:
> Can we add "create collection" in the "select collection" drop down in the nft creation form as well?

+1 

plus we have a hole in this. When a user connects its wallet and then proceed to the page, he can still disconnect 
- for this, could you please make the button label "please connect your wallet" 
image

- and one small detail, 16px text please

components/create/CreateLanding.vue
Jarsen136: >Why don't you keep nuxt-link and use @click.prevent?

It's not allowed
image

In this case, it's `<div>` with a click event handler.

components/create/CreateLanding.vue
Jarsen136: `navigateTo` works as the same

Jarsen136:
> > Can we add "create collection" in the "select collection" drop down in the nft creation form as well?
> 
> +1
> 
> plus we have a hole in this. When a user connects its wallet and then proceed to the page, he can still disconnect
> 
> * for this, could you please make the button label "please connect your wallet"
>   image
> * and one small detail, 16px text please

✅ Done

I have also unified the "select collection" process in both `massmint` and  `nft creation` so that we could delete some duplicate code.

daiagi (approved):
Tested on mobile with nova wallet
Wfm

Maybe move `chooseCollectionDropdown` outside of massmint? 

Other than that lgtm

Jarsen136:
> Tested on mobile with nova wallet Wfm
> 
> Maybe move `chooseCollectionDropdown` outside of massmint?
> 
> Other than that lgtm

Sure. I have moved it. ✅

prury:
Showing Create Nft instead of Creating Collecion: 
image

Jarsen136:
> Showing Create Nft instead of Creating Collecion

fixed

exezbcz:
one last detail, could you please remove the drop shadow from the selector - keeping the rule of no double drop shadow
image

prury:
why do you fail my dear playwright, even if the Not Enough Funds message is still there

Jarsen136:
> why do you fail my dear playwright, even if the Not Enough Funds message is still there

I fixed it. It's because the message changes to  "please connect your wallet" when a user does not log in. I have added a login process to this e2e test case, and then it works.

Jarsen136:
> one last detail, could you please remove the drop shadow from the selector - keeping the rule of no double drop shadow image

✅ Done

components/create/CreateLanding.vue
Jarsen136: Do you mean making such a change as in my screenshot? It does not work after I take a try and cause two duplicate url on the same element.
image

Even though I have tried another solution, it is still the same. https://stackoverflow.com/questions/59885664/nuxt-how-to-prevent-nuxt-link-goes-to-another-page

So, I would keep the original implementation in my PR. If you have any other ideas, feel free to let me know.

roiLeo: noted, I'll keep you informed

prury:
> > why do you fail my dear playwright, even if the Not Enough Funds message is still there
> 
> I fixed it. It's because the message changes to "please connect your wallet" when a user does not log in. I have added a login process to this e2e test case, and then it works.

oh my friend, thank you so much!

Issue #7375, opened 2023-09-26
Nuxt3 - Encountered problems list


prury:
### What happened?

- teleport page pending
- transfer button to change from usd to token not there
- collection and nft previews won't show when creating nft/collection
- list nft for sale switch doesn't hide show input
- footer subscribe input/button pending

### Please reproduce in steps

a

### Expected Behavior

a

### Are you logged in?

None

### Screenshots

a

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

roiLeo:
@kodadot/external-contributors & @kodadot/code-review-guild if you guys want to help prury made a list of thing that is not working on `nuxt` branch.

I would also add: (from https://github.com/kodadot/nft-gallery/issues/7034#issuecomment-1736882919)

- [ ] consola plugin
- [ ] test transaction 
- [ ] download original item
- [ ] playwright tests on deployement
- [ ] fix useVModel in `BasicTokenForm`
- [ ] replace $seoMeta, some of theme have been made in #7297

lmk which one you want to work on

hassnian:
fixes : #7422

-   transfer is not showing address checking messages
-  transfer page total not summing values

yangwao:
- https://github.com/kodadot/nft-gallery/issues/7519

preschian:
Sometimes, I can’t switch pages. There’s an intermittent error. It only happens in the production build. In development mode, I never get an error like that. The cause of the error is unknown, and how to reproduce it has not yet been determined. When reloaded, the error disappears.

roiLeo:
> Does anyone face that error also on nuxt.kodadot.xyz?

Yes! page frozen, but route change.
 I didn't manage to reproduce it on local env too.

vikiival:
> Yes! page frozen, but route change.

Guess some particular overfetch ? 
it is highly possible that too many requests / awaits / something will freeze the app

roiLeo:
> Guess some particular overfetch ? it is highly possible that too many requests / awaits / something will freeze the app

When navigate through https://nuxt.kodadot.xyz/ahp/gallery/2-2 for example

newraina:
- [ ] https://nuxt.kodadot.xyz/snek/incomingoffers page is blank [screenshot](https://github.com/kodadot/nft-gallery/assets/10708802/d6f2d2bc-a437-4585-b9b5-2e6ca1b7ed17)

preschian:
> Yes! page frozen, but route change.
[...]

PR #7575, opened 2023-10-10
🔧 GalleryItem SEO

roiLeo:
- ref #7375 

Tested on `/ksm/gallery/16759320-8453d0ccb4cb7e9e59-A51-THE_FOUNDRY-00000246`

Capture d’écran 2023-10-10 à 10 28 48 AM

### DIFF
--- a/components/gallery/GalleryItem.vue
+++ b/components/gallery/GalleryItem.vue
@@ -266,27 +266,28 @@ onMounted(() => {
 const { isUnlockable, unlockLink } = useUnlockable(collection)
 
 const title = computed(() => nftMetadata.value?.name || '')
-const meta = computed(() => {
-  return [
-    {
-      title: title.value,
-      description: convertMarkdownToText(nftMetadata.value?.description),
-      image: generateNftImage(
-        title.value,
-        formatBalanceEmptyOnZero(nft.value?.price as string),
-        sanitizeIpfsUrl(nftImage.value || ''),
-        nftMimeType.value,
-      ),
-      mime: nftMimeType.value,
-      url: route.path,
-      video: sanitizeIpfsUrl(nftAnimation.value || ''),
-    },
-  ]
+const seoDescription = computed(
+  () => convertMarkdownToText(nftMetadata.value?.description) || '',
+)
+const seoCard = computed(() => {
+  if (nft.value) {
+    return generateNftImage(
+      title.value,
+      formatBalanceEmptyOnZero(nft.value?.price as string),
+      sanitizeIpfsUrl(nftImage.value || ''),
+      nftMimeType.value,
+    )
+  }
 })
 
-useHead({
+useSeoMeta({
   title,
-  meta,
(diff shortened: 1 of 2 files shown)


roiLeo:
idk why but I can see the metadata in my plugin

Capture d’écran 2023-10-10 à 10 41 13 AM

preschian (approved):

preschian:
> idk why but I can see the metadata in SeoMetaInspector

it means we still need [ssr-opengraph](https://github.com/kodadot/workers/blob/main/ssr-opengraph/wrangler.toml) or [prerender](https://github.com/kodadot/workers/blob/main/prerender/wrangler.toml) workers

Unless we found a way to enable ssr mode on our `nuxt` branch. https://nuxt.com/docs/api/composables/use-server-seo-meta

currently nuxt subdomain is served on netlify. I can't put ssr-opengraph or prerender on netlify

Screenshot 2023-10-10 161846

roiLeo:
> it means we still need [ssr-opengraph](https://github.com/kodadot/workers/blob/main/ssr-opengraph/wrangler.toml) or [prerender](https://github.com/kodadot/workers/blob/main/prerender/wrangler.toml) workers

Do we need to run a worker to fix this?

> Unless we found a way to enable ssr mode on our `nuxt` branch. https://nuxt.com/docs/api/composables/use-server-seo-meta

We could try https://github.com/harlan-zw/nuxt-og-image & `nuxt-seo-kit` package

> currently nuxt subdomain is served on netlify. I can't put ssr-opengraph or prerender on netlify

do you mean that it need to be on the same subdomain to generate image?

preschian:
> Do we need to run a worker to fix this?
> 
> > Unless we found a way to enable ssr mode on our `nuxt` branch. [nuxt.com/docs/api/composables/use-server-seo-meta](https://nuxt.com/docs/api/composables/use-server-seo-meta)
> 
> We could try [harlan-zw/nuxt-og-image](https://github.com/harlan-zw/nuxt-og-image) & `nuxt-seo-kit` package

if we can enable SSR support in the `nuxt` branch, it's better to drop prerender workers. use that nuxt-og-image or nuxt-seo-kit instead

yangwao:
I would vote for nuxt-og-image as we can play with that easier

roiLeo:
> if we can enable SSR support in the `nuxt` branch, it's better to drop prerender workers. use that nuxt-og-image or nuxt-seo-kit instead

I'm pretty sure that if we enable ssr support now it will break many things (`window` & `document` need to be on client-side)

preschian:
> > if we can enable SSR support in the `nuxt` branch, it's better to drop prerender workers. use that nuxt-og-image or nuxt-seo-kit instead
> 
> I'm pretty sure that if we enable ssr support now it will break many things (`window` & `document` need to be on client-side)

yes 👍

I think if we want to start to check ssr mode, we can start from `/blog` first. In the meantime, we still use our prerender workers.

Issue #7625, opened 2023-10-12
Put one AHP collection in sitemap.xml


yangwao:
![image](https://github.com/kodadot/nft-gallery/assets/5887929/6573a0b9-cc07-40eb-ba46-0b782a80374b)

https://kodadot.xyz/sitemap.xml -> https://kodadot.xyz/sitemap_index.xml 

and then to https://kodadot.xyz/0-sitemap.xml 

~~is this legit?~~

seems yes, google discovered finally new things, let's see if that give us ups or downs

![image](https://github.com/kodadot/nft-gallery/assets/5887929/827b4f54-80b5-4320-bb6b-41d2e290c21c)

Can I have in sitemap link to one ahp collection for testing purposes https://kodadot.xyz/ahp/collection/13 pls

PR #7626, opened 2023-10-12
🧪 sitemap collection

roiLeo:
- ref #7625

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -284,6 +284,10 @@ export default defineNuxtConfig({
     // },
   },
 
+  routeRules: {
+    '/ahp/collection/13': { sitemap: { changefreq: 'daily', priority: 0.3 } },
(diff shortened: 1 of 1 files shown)


nuxt.config.js
```diff
@@ -284,6 +284,10 @@ export default defineNuxtConfig({
     // },
   },
 
+  routeRules: {
+    '/ahp/collection/13': { sitemap: { changefreq: 'daily', priority: 0.3 } },
+  },
+
```
Jarsen136, line 290: It could also work.

image

Jarsen136 (approved):

nuxt.config.js
roiLeo: Hmmmmm nice I'll look Into that

preschian (approved):

PR #7725, opened 2023-10-18
🔧 POW Server side rendering

roiLeo:
Let me know your thoughts about this change.

SSR is needed if we want to use all Nuxt power


roiLeo:
😢 
Screenshot 2023-10-18 at 11-50-40 Invocation Failed

preschian:
as long the build and deployed successfull im fine with it 👍🏻

roiLeo:
So many things to reconfigure on client-side, don't mind first load only showing footer it's normal.
Looks like all playwright tests are broken now 😞

yangwao:
What can potentially break?

roiLeo:
> What can potentially break?

UI/UX, client side plugin and maybe more?

yangwao:
> UI/UX, client side plugin and maybe more?

And do we need to setup something special from CF/Netlify to work?

roiLeo:
> And do we need to setup something special from CF/Netlify to work?

Nope, only check which components needs to be loaded on client-side and which can be loaded on server side.
For now I've wrapped most components on client-side that why we have a "delayed loading"

roiLeo:
// TODO: persist localstorage (or maybe vueuse equivalent)

roiLeo:
more changes in #8356

PR #7763, opened 2023-10-21
fix: sort is not always an array

daiagi:

## PR Type

- [x] Bugfix

### <samp>🤖 Generated by Copilot at 9b92bd8</samp>

Fixed a bug in `useChainRedirect.ts` that caused an invalid sort parameter to be passed to the API when redirecting from a remark prefix. Added a type check and a conversion logic for the `restOfQuery.sort` value.

### <samp>🤖 Generated by Copilot at 9b92bd8</samp>

> _`restOfQuery.sort`_
> _Array or string? Check first_
> _Autumn leaves no bugs_


composables/useChainRedirect.ts
```diff
@@ -67,9 +67,13 @@ function getRedirectPathForPrefix({
...
     if (!isRemark.value && restOfQuery.sort) {
-      restOfQuery.sort = restOfQuery.sort.filter(
-        (value) => !value.startsWith('instance_'),
-      )
+      if (Array.isArray(restOfQuery.sort)) {
+        restOfQuery.sort = restOfQuery.sort.filter(
+          (value) => !value.startsWith('instance_'),
+        )
+      } else if (restOfQuery.sort.startsWith('instance_')) {
+        delete restOfQuery.sort
```
roiLeo, line 75: so are we using delete or not?

roiLeo (approved):
this part is still not easy to read, maybe we should refactor it or use better naming

daiagi:
>this part is still not easy to read, maybe we should consider refactoring or use better naming

you're right.
please have a look again (although you approved already)

preschian (approved):

prury:
i still get error tho, not sure if is the same error because it does not show any info, but is in the exact same line: 

image

daiagi:
>i still get error tho, not sure if is the same error because it does not show any info, but is in the exact same line:


not the same error


seems to be metamask related:

metamask disabled:

image


enabled:

image

prury:
> > i still get error tho, not sure if is the same error because it does not show any info, but is in the exact same line:
> 
> not the same error
> 
> seems to be metamask related:
> 
> metamask disabled:
> 
> image
> 
> enabled:
> 
> image



> > i still get error tho, not sure if is the same error because it does not show any info, but is in the exact same line:
> 
> not the same error
> 
> seems to be metamask related:
> 
> metamask disabled:
> 
> image
> 
> enabled:
> 
> image

ok, thank you and sorry for the trouble

PR #7758, opened 2023-10-20
⬆️ nuxt version

roiLeo:
👷‍♂️ 
- close #7754 
- remove unused stuff

PR #7764, opened 2023-10-21
pnpm update nuxt

yangwao:
does it build

continue from
- #7758


yangwao:
builds, thanks to @roiLeo !

Issue #7562, opened 2023-10-09
Migration UI - 3 - Signing 


exezbcz:
part of a bigger issue:
- #7552 

# Signing 🚥 
- new part, will be the same as with auto teleport (but in modal) 
- the reason everything is on its own page and not in the small modal is simply the space and no way of accidentally closing it and losing the progress. It's quite a lot of transactions if you decide to migrate 1k collection. 
<img width="373" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/c2f880ee-69e6-444f-a633-07c68cbecaf3">

- it's a step signing.
- here are possible events of the small sub-part 
<img width="572" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/f17064f3-8e42-4f4e-be46-031199365796">

- the parent section changes based on whether its in progress, its done, or inactive, changes between three states. 
<img width="119" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/e01e518a-6775-4689-8d61-4222f94aa006"> <img width="71" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/40ca3d0a-6073-45c6-92bc-f158cdc5435d"> <img width="120" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/bfbacaba-80c5-49d2-8155-74ab9fbf8515">

- the grey 1/4 circle is a loading wheel, please match the style as much as possible. 
- The same icons are used for the subsections. Every step is in this case one signed transaction. 
- let me give an example. 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/16f8397b-4f07-4a91-bddf-91607fc9a3f4)

- the contextual message is in this case what is changing. 
- when you cancel signing, you have the option to retry (it could say failed or canceled) 
<img width="627" alt="image" src="https://github.com/kodadot/nft-gallery/assets/90852205/5da95aa4-8e6b-4ea3-8eed-d280b3f565b0">

- after you finish all of this long signing, you can click on the button to finish (it's not automatic because in this signing, there are also transaction links (as you can see above)

exezbcz:
Some changes will be made, mainly to the structure of what you are signing, and maybe some flow as well. 

don't take this issue please till we update that. 

cc @vikiival

yangwao:
@exezbcz  where the user will see if everything was migrated successfully with on-chain checks?

exezbcz:
[...]

PR #7908, opened 2023-11-01
feat: assethub show empty collection and delete collection

### LINKED ISSUE #4420: Making empty collections visible to creator
### Is your feature request related to a problem?

When someone creates a collection and will not add NFT right after into it, this collection is not visible in his profile. I just chatted with a confused artist who wanted to report an issue because he could not find his (empty) collection. This is probably not an essential change, but it would make our platform a bit more user-friendly. 

Artists can see their empty collection only in the process of creating new NFTs.

### Describe the solution you would like

Maybe it will be nice to have something like drafts or empty collections that would be visible only to the artist. 

### Describe alternatives you have considered

Add a folder or option where people can see their empty collections.

### Screenshots

_No response_

### Are there opened related issues?

_No response_

### LINKED ISSUE #7553: Make empty colletions visible for the creator
This is not the first time someone is confused because they cannot see their empty collection in their profile. At the moment if you have empty collection you can only see it in creation process when you add your NFTs to existing collection.
Screenshot 2023-10-08 at 16 14 05

Over time, there were more people mentioning this. We also have an issuefor this but I want to bring back attention to this because no one picked it up. 
- https://github.com/kodadot/nft-gallery/issues/4420

We need to let people know that their empty collection exists (not only when they want to add NFTs to it)

**Suggestions:**
- make empty collections visible only to creators
- add a filters to profile (empty collections). Somewhere here:
Screenshot 2023-10-08 at 16 18 35 

- add somewhere text: empty collections will be visible once you upload your first nft to them somewhere in to profile
- idk something so people are not confused about this

preschian:

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #7553
- [x] Closes #4420
- [x] I need this, to easily delete empty collection while testing #7562

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1xjvRADwdJcnmUCLWerEHR4iGCT5EBTm4D4fzLLg4LcAC9p)

<details><summary>Screenshot</summary>
<p>

image

</p>
</details> 

### <samp>🤖 Generated by Copilot at 2187942</samp>

This pull request adds and modifies some components and queries to enable showing and deleting collections on profile pages for AssetHub and Basilisk chains. It uses a new GraphQL query file `collectionListWithSearchProfile.graphql` to fetch collections by current owner, and updates `CollectionGrid.vue` and `HeroButtons.vue` to handle different chains and actions.

### <samp>🤖 Generated by Copilot at 2187942</samp>

> _Sing, O Muse, of the skillful coder who wrought_
> _Many changes on the web app of collections, where_
> _Users can display and trade their digital treasures, and_
> _Invoke the power of GraphQL to fetch and filter them._


components/collection/HeroButtons.vue
```diff
@@ -130,6 +132,31 @@ const QRModalActive = ref(false)
...
+      async ({ status }) => {
+        if (status.isInBlock) {
+          toast($i18n.t('Collection Deleted'))
+
+          await delay(DETAIL_TIMEOUT)
+          navigateTo(`/${urlPrefix.value}/u/${accountId.value}?tab=collections`)
+        }
+      },
+    )
+}
```
daiagi, line 159: move to composable transaction?

prury:
empty collections are showing, but the card information is empty on all of them: 
image

preschian:
@daiagi @prury updated, how about now

components/collection/HeroButtons.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div>
+    <Loader v-model="isLoading" :can-cancel="false" />
```
daiagi, line 3: Loader in HeroButtons?
a bit strange

components/collection/HeroButtons.vue
```diff
@@ -130,6 +134,48 @@ const QRModalActive = ref(false)
...
+    query: collectionByIdMicro,
+    clientId: client.value,
+    variables: {
+      id: id.toString(),
+    },
+  })
+
+  if (!isLoading.value) {
+    return
+  }
```
daiagi, line 158: move this above the `collectionByIdMicro` to avoid premature query

components/collection/HeroButtons.vue
```diff
@@ -130,6 +134,48 @@ const QRModalActive = ref(false)
...
+
+  if (data.value.collectionEntity?.burned) {
+    navigateTo(`/${urlPrefix.value}/u/${accountId.value}?tab=collections`)
+  } else {
+    await delay(DETAIL_TIMEOUT)
```
daiagi, line 163: can we avoid the delay?
maybe by watching something and triggering the checkCollection function at the right time?

queries/subsquid/general/collectionListWithSearchProfile.graphql
```diff
@@ -0,0 +1,36 @@
...
+      metadata
+      name
+      price
+      burned
+      currentOwner
+      events(where: { interaction_eq: BUY }) {
+        meta
+      }
+    }
+  }
```
daiagi, line 29: can the query be slimmed down?
do we need the nfts?
do we need the events?

daiagi (changes requested):

components/collection/HeroButtons.vue
preschian: If there is no loader once the owner clicks the delete button nothing happens. It will lead to bad UX, I guess

daiagi: i get the need for the Loader
just think it is more suitable to be at the page level and not in a sub-component

up to your consideration, if you find it is not worth the hustle then I let it go

components/collection/HeroButtons.vue
preschian: Hmm, any idea?

Because I can't rely on `blocknumber`. To make sure the collection was burned, then navigate to the user collection page

queries/subsquid/general/collectionListWithSearchProfile.graphql
preschian: yes, to fix @prury issue

components/collection/HeroButtons.vue
daiagi: `/composables/useSubscriptionGraphql`  maybe?

components/collection/HeroButtons.vue
preschian: ahh I see. agree on that, would be better if I can call something like global loader

components/collection/HeroButtons.vue
preschian: checking

components/collection/HeroButtons.vue
daiagi: yeah.
I had a similar situation with the shopping cart, where the loader could be triggered from anywhere in the app

that's why I created the <Buy/> component

components/collection/HeroButtons.vue
preschian: Thanks, it looks better with subscriptions. I forgot that we had that composable

components/collection/HeroButtons.vue
preschian: Temporarily, I moved it to the layout. It would be good if we could put it somewhere globally

components/collection/CollectionGrid.vue
```diff
@@ -95,6 +97,9 @@ onBeforeMount(() => {
 })
 
 const fetchPageData = async (page: number, loadDirection = 'down') => {
+  const isProfilePage =
+    route.name === 'prefix-u-id' && (isAssetHub.value || isBasilisk.value)
+
```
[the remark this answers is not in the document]
preschian: updated, looks fine without computed. and adjusted on remark

composables/transaction/collection/delete.ts
```diff
@@ -0,0 +1,29 @@
+import type { ApiPromise } from '@polkadot/api'
+import type {
+  ActionDeleteCollection,
+  ExecuteTransactionParams,
+} from '@/composables/transaction/types'
+
+import { warningMessage } from '@/utils/notification'
+
+export function execDeleteCollection(
```
roiLeo, line 9: or move it inside `transactionBurn` composable?

layouts/explore-layout.vue
```diff
@@ -36,6 +36,8 @@
       <ListingCartMini />
       <ListingCartModal />
     </template>
+
+    <Loader v-model="isLoading" :can-cancel="false" />
```
roiLeo, line 40: hmm.. any side effects with child component that already import this Loader?
I would like to keep logic outside `layouts`

roiLeo (approved):
otherwise lgtm

layouts/explore-layout.vue
daiagi: thinking same
could cause double loaders

components/collection/CollectionGrid.vue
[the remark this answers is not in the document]
daiagi: didn't fully understand the remark logic of remove the "burned" param
but I trust it's for a good reason

daiagi (approved):
agree with @roiLeo  about the Loader
otherwise okie

prury:
> @daiagi @prury updated, how about now

I was sure i had edited my comment yesterday with one more thing, but it seems github was sort of bugged. 

Minor thing: the collection counters don't sum empty collections

composables/transaction/collection/delete.ts
preschian: updated

layouts/explore-layout.vue
preschian: > hmm.. any side effects with child component that already import this Loader?
> I would like to keep logic outside `layouts`

I think it should be fine because they use a different `v-model` source. I moved the state and update state loader on the layout to the plugin here https://github.com/kodadot/nft-gallery/blob/feat/show-empty-collection/plugins/loader.ts



> thinking same
> could cause double loaders

As long as the child component is not set to multiple loader state, it should be fine. Better if the child component uses `$updateLoader` instead

Please note, currently I only put it on one layout. Would be better if we put a loader on all layouts

preschian: child component should be fine I guess, can continue with this:
```vue
<template>
  <div>
    <Loader v-model="isLoading" />
    <button @click="toggleLoader()">Toggle Loader</button>
  </div>
</template>

<script lang="ts" setup>
const isLoading = ref(false)

const toggleLoader = () => {
  isLoading.value = !isLoading.value
}
</script>
```

it's better to change the child component to this:
```vue
<template>
  <div>
    <button @click="toggleLoader()">Toggle Loader</button>
  </div>
</template>

<script lang="ts" setup>
const { $loader, $updateLoader } = useNuxtApp()

const toggleLoader = () => {
  $updateLoader(!$loader.value)
}
</script>
```

preschian:
> Minor thing: the collection counters don't sum empty collections

@prury updated, how about now

prury:
> > Minor thing: the collection counters don't sum empty collections
> 
> @prury updated, how about now

thank you!

yangwao:
Thanks!
pay 50 usd

PR #7916, opened 2023-11-01
fix: Mint phase time is off for generative drop

### LINKED ISSUE #7884: Mint phase time is off
Drop page: Our latest generative drop on ahk. 

Once one hour passes, minting time resets to anouther hour.. and again and again:

image

image

It made sense when we had limited mints per hour (before we had mint limited to 10 mints per hour). Now it does not make sense to have time that reserts every hour. 

From a marketing perspective:
- its confusing
- feels limiting but its not
- I felt bamboozled when I realized it
- if the limit of the campaign is a number of the NFTs max minted (200) then turn of the time counter

Solution:
- limit campaing on a real time of campaing - lets say one week: the time will show reality when the campaign ends. 
- campaign with no limitation (time/number) does not feel inclusive or rare
- as I said above: if the limit of the campaign is a number of the NFTs max minted (200) then turn of the time counter and dont confuse people

Jarsen136:
## PR Type

  - [x] Bugfix

  ## Needs Design check

  - @exezbcz please review

  ## Context

  - [x] Closes #7884 

  ## Screenshot 📸

  - [x] My fix has changed UI
image

### DIFF
--- a/components/collection/drop/Generative.vue
+++ b/components/collection/drop/Generative.vue
@@ -1,8 +1,7 @@
 <template>
   <div class="unlockable-container">
     <CollectionUnlockableLoader v-model="isLoading" :minted="justMinted" />
-    <CountdownTimer />
-    <div class="container is-fluid">
+    <div class="container is-fluid border-top">
       <div class="columns is-desktop">
         <div class="column is-half-desktop mobile-padding">
           <UnlockableCollectionInfo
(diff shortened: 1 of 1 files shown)


prury:
you just removed the counter? it works for me! hehe
we need more gen art, 80 mints already??

prury (approved):

yangwao:
thanks!
pay 20 usd

yangwao (approved):

Issue #7826, opened 2023-10-25
Current Testing State


prury:
# *Still editing*

| **Test Page** | **Priority** |
| ------------- | ------------- |
Assets		| wait for future refactor
Basic | ✔️ 
Blog		|  p5
Create Collection		|  ✔️ 
Create NFT		|  p1 
Cart | 
Collection | p1
CSS		|  p5
Drops	|  p2
Explore 		|  ✔️ (@nick-dev123 )
Footer		|  ✔️ 
Gallery Item		|  ✔️
Identity		|  p3
Landing		|  under development
Language Change		|  ✔️ 
Mass listing |  needs account solution
Navbar | p2
Notifications		| p5
Open graph | needs fixing by:
Prefixes | ✔️ 
Profile | ✔️ 
Profile Sidebar		| p1
Redirection | ✔️ 
Search		|  ✔️ 
Shopping Cart		| p3
Sort		|  p3
Transfer		|  ✔️ 
Teleport Bridge | p2
Transfer NFT		|  p3
Wallet Interactions		| p1

**Basic:** 
- verify element existence on landing page across all chains

**Create Collection:** 
- upload image to drag and drop
- Verify collection name input existence and fill it
- fill collection description input
- unlimited items switch activation
- verify error message on unlimited items switch
- create collection button existence and if its deactivated
- verify if deposit amount and collection balance loads properly 
- verify if warning message appears if user did not filled name for collection

**Explore:**

**Footer:**
- Cycle trough all footer links and check if they provide correct address
- Cycle trough all footer social media links and check if they provide correct address
- Check if blog link redirects to internal blog page
- Check footer subscription existence and if it throws error if incomplete e-mail address is filled

**Prefixes:**
- Cycle trough all chains verifying if the proper name shows on navbar chain drop down
- Verifies if clicking on an RMRK2 NFT redirects to proper address
- Verifies if clicking on an Ahk NFT redirects to proper address

**Redirect:**
- Checks if using a /rmrk2 prefix on a collection redirects properly to /ksm
- Checks if using a /rmrk2 prefix on a item redirects properly to /ksm

kodabot:
ASSIGNED - @prury 🔒 LOCKED -> Thursday, October 26th 2023, 15:46:40 UTC -> 24 hours

vikiival:
I would replace BSX with AssetHubs as they have higher prio ^-^

roiLeo:
[...]

PR #7849, opened 2023-10-27
Enviroment var to expand test cases

prury:
Added a fixed mnemonic inside a Github Secret variable to expand test cases.
Before, we had a random mnemonic without balance or nfts in it, making it impossible to test confirmation modals, mass minting, shopping cart and masslisting. 

Tests are not probably not passing because of this:
image
or this: 
image

But as you can see, a copy of this PR on my forked repository with the same secret set works just fine: 
[Prury nft-gallery](https://github.com/prury/nft-gallery/pull/1)

## PR Type

- [x] Testing feature

## Context

- [ ] Related to: #7826
- [ ] Requires deployment <snek/rubick/worker>

### <samp>🤖 Generated by Copilot at 568beaf</samp>

This pull request enables the use of a secret mnemonic phrase for end-to-end testing, and updates the e2e-login page and the test cases accordingly. It also fixes a minor bug in the createcollection test.

### <samp>🤖 Generated by Copilot at 568beaf</samp>

> _We use the secret phrase to enter the realm_
> _We test the account with the avatar's helm_
> _We create new collections with our funds_
> _We fix the broken tests with our guns_

PR #7938, opened 2023-11-03
Test: Create NFT page

prury:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [x] Test

## Context

- [x] Related #7826 

### <samp>🤖 Generated by Copilot at c79c41e</samp>

This pull request adds `data-testid` attributes to various components in the create nft feature to enable automated testing with Playwright. It also adds a new test file and a new method to the Commands class to test the creation of an nft and the opening of a new tab.

### <samp>🤖 Generated by Copilot at c79c41e</samp>

> _We are the testers of the nft world_
> _We use Playwright to unleash our wrath_
> _We add data-testid to every component_
> _We make them scream with our assert_


tests/e2e/createnft.spec.ts
```diff
@@ -0,0 +1,95 @@
+import { expect, test } from './fixtures'
+
+const prefix = ['ahk', 'ahp', 'ksm']
```
daiagi, line 3: why  have an array if you only use 'ksm'

daiagi (approved):
looks good
maybe it is too much to ask, but it would be good to also test that the transaction is sent with the correct details
as if you were looking on the args in the wallet signing modal

prury:
> looks good maybe it is too much to ask, but it would be good to also test that the transaction is sent with the correct details as if you were looking on the args in the wallet signing modal

yeah that would be great, but for this to be possible, we either have to merge #7849 or have a proper testnet, because i cannot proceed to the screen prior to the signing phase without balance on the account

tests/e2e/createnft.spec.ts
prury: removed

yangwao:
> because i cannot proceed to the screen prior to the signing phase without balance on the account

wyt about setting some "dev" environment variable for codebase 👀

prury:
> > because i cannot proceed to the screen prior to the signing phase without balance on the account
> 
> wyt about setting some "dev" environment variable for codebase 👀

using a dev seed? i think the limitations would be the same, no or low balance, no nfts/collections in it

preschian (approved):

PR #7905, opened 2023-10-31
Fix : Identity creation: Add Select chain #7823

AshutoshSingh00001:
Thank you for your contribution to the [KodaDot - One Stop Shop for Polkadot NFTs](https://kodadot.xyz/).

👇 __ Let's make a quick check before the contribution.

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #7823
- [x] ## Did your issue had any of the "$" label on it?
- [x] My DOT address: [Payout](https://kodadot.xyz/dot/transfer?target=158VTDxZJidzsMcAdh7h828domveRvpmk1SCxwLUEHhcgY4v&usdamount=0&donation=true)

PR #7944, opened 2023-11-04
Feat: Add Select chain #7823 #6483

### LINKED ISSUE #6483: Statemine identity creation not working
I tried to create my identity on statemine (production and canary), and I got stuck both times here:

Screenshot 2023-07-28 at 16 22 08

### LINKED ISSUE #7823: Identity creation: Add Select chain
It would be cool to have an option to select on what chain you want to create your identity. At the moment its possible only  change it only when you change chain in the main menu.

Screenshot 2023-10-25 at 12 10 03

AshutoshSingh00001:
Thank you for your contribution to the [KodaDot - One Stop Shop for Polkadot NFTs](https://kodadot.xyz/).

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #7823
- [x] Closes #6483
- [x] ## Did your issue had any of the "$" label on it?
- [x] My DOT address: [Payout](https://kodadot.xyz/dot/transfer?target=158VTDxZJidzsMcAdh7h828domveRvpmk1SCxwLUEHhcgY4v&usdamount=0&donation=true)


AshutoshSingh00001:
@roiLeo recreated pr #7905 because of too many commits in previous one

components/common/IdentityForm.vue
```diff
@@ -288,6 +323,14 @@ const hasIdentity = computed(() => {
   )
 })
 
+let Createidentity
+
+if (hasIdentity.value) {
+  Createidentity = 'Update Identity'
```
vikiival, line 329: Add to translation files

components/common/IdentityForm.vue
```diff
@@ -288,6 +323,14 @@ const hasIdentity = computed(() => {
   )
 })
 
+let Createidentity
```
vikiival, line 326: Never let

components/common/IdentityForm.vue
```diff
@@ -288,6 +323,14 @@ const hasIdentity = computed(() => {
...
+
+if (hasIdentity.value) {
+  Createidentity = 'Update Identity'
+} else {
+  Createidentity = 'Create Identity'
```
vikiival, line 331: We use camelCase

components/common/IdentityForm.vue
```diff
@@ -227,6 +242,26 @@ const {
 const isConfirmModalActive = ref(false)
 const isLoaderModalVisible = ref(false)
 const transactionValue = ref('')
+const menus = availablePrefixes().filter(
+  (menu) =>
```
vikiival, line 246: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

vikiival (changes requested):

components/common/IdentityForm.vue
AshutoshSingh00001: done added identity.update

components/common/IdentityForm.vue
AshutoshSingh00001: replaced this with v-if="!hasIdentity" within button

components/common/IdentityForm.vue
AshutoshSingh00001: removed this line of code

components/common/IdentityForm.vue
AshutoshSingh00001: well I tried to figure out issue but I didn't found any bugs with this line of code can you explain me more about this @vikiival

components/common/IdentityForm.vue
AshutoshSingh00001, line 1: ok now I think I got your point you mean inn such a case someone chooses ahk or ahp from nav menu then there should be a way to restrict them and we already have it here

AshutoshSingh00001 (commented):
ok now I think I got your point you mean inn such a case someone chooses ahk or ahp from nav menu then there should be a way to restrict them and we already have it here

daiagi:
on Firefox
choosing network from the dropdown causes the page to look like this:
image

AshutoshSingh00001:
> on Firefox choosing network from the dropdown causes the page to look like this: image

fixed recheck

components/common/IdentityForm.vue
AshutoshSingh00001: done fixed if isAssetHub Create or Update button will get disabled
I tried a different way :)

AshutoshSingh00001:
@vikiival my pr is accordingly  ready

AshutoshSingh00001:
@vikiival @roiLeo waiting for review

components/common/IdentityForm.vue
```diff
@@ -31,6 +31,21 @@
         <hr class="my-7" />
       </div>
 
+      <NeoField :label="$t('Select Blockchain')">
+        <div class="w-100">
```
daiagi, line 35: this `<div>` is redundant

AshutoshSingh00001: yeah I was also thinking about removing it but unfortunatelyi skipped it thanks for the correction

AshutoshSingh00001:
@daiagi are we finally able to merge this one

components/common/IdentityForm.vue
roiLeo: code is still present

AshutoshSingh00001: > code is still present

yeah and it should be there vikival wanted to disable that create identity button because when users select ahp or ahk from nav it used to activate Create button and to disable that he recommended me a way to filter then disable button using javascript but I tried a different way using isAssetHub disabled within Neobutton

in this code it is disabling asset hub and moonriver, moonbeam in the selection
so we are just keep disabling them so users don't get stuck around disabled button after selecting one of these chains :)

AshutoshSingh00001: > code is still present

I'm thinking about user experience if you think we should remove it I'm ready
as you know currently ahk and ahp don't support on-chain identity so adding chain over there won't be good idea once both of them supports identity then we can enable these chains

components/common/IdentityForm.vue
```diff
@@ -101,10 +114,21 @@
...
+        :loading="isLoading"
+        expanded
+        @click="openConfirmModal" />
+
+      <NeoButton
+        v-if="hasIdentity"
+        class="is-flex is-flex-grow-1 fixed-height"
+        variant="k-accent"
+        :label="$t('identity.update')"
+        :disabled="disabled || isAssetHub"
```
roiLeo, line 131: doesn't need 2 button, you can only use one

components/common/IdentityForm.vue
```diff
@@ -227,6 +251,26 @@ const {
 const isConfirmModalActive = ref(false)
 const isLoaderModalVisible = ref(false)
 const transactionValue = ref('')
+const menus = availablePrefixes().filter(
+  (menu) =>
+    menu.value !== 'movr' &&
+    menu.value !== 'glmr' &&
+    menu.value !== 'ahk' &&
+    menu.value !== 'ahp',
```
roiLeo, line 259: consider refactoring

components/common/IdentityForm.vue
```diff
@@ -227,6 +251,26 @@ const {
...
+const selectChain = ref(chainByPrefix.value?.value || menus[0].value)
+watch(urlPrefix, (value) => {
+  selectChain.value = value
+})
+const currentChain = computed(() => selectChain.value as Prefix)
+watch(currentChain, () => {
+  if (currentChain.value !== urlPrefix.value) {
+    setUrlPrefix(currentChain.value as Prefix)
+  }
+})
```
roiLeo, line 273: I'll put the computed & watcher together

roiLeo (commented):
- when changing wallet it doesn't update the handle input
- can't set/update different identity between chains it's always the same

components/common/IdentityForm.vue
AshutoshSingh00001: done

components/common/IdentityForm.vue
AshutoshSingh00001: done

AshutoshSingh00001:
> * when changing wallet it doesn't update the handle input
> * can't set/update different identity between chains it's always the same

clear form on change

components/common/IdentityForm.vue
AshutoshSingh00001: > I'll put the computed & watcher together

how

AshutoshSingh00001: > I'll put the computed & watcher together

you mean for currentChain

AshutoshSingh00001: obviously you can but in my opinion I don't think making changes in this line of code will make a any difference regardless of this you can check other select chain button on create page etc they work similar and there is no any major difference

AshutoshSingh00001:
> * when changing wallet it doesn't update the handle input
> * can't set/update different identity between chains it's always the same

you're right and it works in the similar way because this is saves time and effort suppose when a user has selected wrong chain and then he tries to switch and all of his input clears so with this thought in mind I tried to make it user friendly and not a trouble any more

components/common/IdentityForm.vue
roiLeo: better reading

roiLeo (commented):
Spacing issue:

Screenshot 2023-11-09 at 09-22-15 KodaDot - NFT Market Explorer


> you're right and it works in the similar way because this is saves time and effort suppose when a user has selected wrong chain and then he tries to switch and all of his input clears so with this thought in mind I tried to make it user friendly and not a trouble any more same concept for wallet also but if someone wants to make small changes they can just by reentering new data

add a watcher? I tried on `snek` identity was cleared but it was still displaying my handle.
After reloading all my chain handle where cleared? have you tested it on your side?

vikiival:
Hello @AshutoshSingh00001 
Due to the low code quality, we have decided to close this PR. 

It is highly time consuming for anyone from @kodadot/internal-dev to do a review of non-working, low-priority issue. 

We would be happy to work with your PR
1. When the code will be working
2. The PR is refferencing to a higher prio issue. 

Thanks for the understanding. 🤗😇

AshutoshSingh00001:
> Hello @AshutoshSingh00001 Due to the low code quality, we have decided to close this PR.
> 
> It is highly time consuming for anyone from @kodadot/internal-dev to do a review of non-working, low-priority issue.
> 
> We would be happy to work with your PR
> 
> 1. When the code will be working
> 2. The PR is refferencing to a higher prio issue.
> 
> Thanks for the understanding. 🤗😇

thanks @vikiival I was also putting a lot of time in in particular issue and the issues are not ending so I'll try to resubmit this pr later once I'll fully test it

PR #7954, opened 2023-11-06
Revert "Enviroment var to expand test cases"

prury:
Reverts kodadot/nft-gallery#7849


roiLeo (approved):

prury:
> so which part doesn't work?

it fails to access the github secret containing the environment variable, because it will always run from a forked repo(just a guess)

PR #8124, opened 2023-11-17
feat: Loading modal for actions

### LINKED ISSUE #8061: Loading modal for actions
image

- I decided to create a loading modal, since auto-teleport takes some time before it checks all the funds
- I am not really fan of landing a modal that is not loaded. Having a loader and then getting to a modal that is 100% ready is IMO better UX. 
- indication that the process is underway 
- avoiding sudden rescaling of the modal once something loads 
- prevents users from clicking buttons that do nothing
- This model will be used for every action where this loading happens. 
- updated in figma file: https://www.figma.com/file/JXNtryaNYhY0XJNArzvPQx/Handoff?type=design&node-id=1458%3A19091&mode=design&t=Usu2Htfj9OHID4N1-1

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

changed: buy, list, mint nft , mint colelction

- [x] Closes #8061

- [x] Fill up your DOT address: [Payout](https://kodadot.xyz/dot/transfer?target=13QUj3pZyFNfYj4AM336hRdyLQbevj5H3sR4PKmLEXLdwZhh)

#### Community participation

## Screenshot 📸

CleanShot 2023-11-20 at 09 49 04

long modal, with scroll 

CleanShot 2023-11-20 at 10 29 00

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at d3c236d</samp>

This pull request introduces a new `BoxedModal` component that standardizes the appearance and behavior of the modal components. It also refactors the `ConfirmPurchaseModal.vue` component to use the new `BoxedModal` and to show a loading state based on the user's balances. Additionally, it adds a new loading animation for the dots element and cleans up some code.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at d3c236d</samp>

> _We're trapped in a modal of doom_
> _Waiting for the balances to load_
> _The dots are spinning in a loop_
> _The `BoxedModal` is our only hope_


exezbcz:
i know its a draft but the close x-mark icon does not work on both modals (loading, loaded) 

image

components/shared/modals/BoxedModal.vue
```diff
@@ -0,0 +1,165 @@
+<template>
+  <div class="modal-width">
+    <header
+      class="px-6 is-flex is-justify-content-space-between border-bottom is-align-items-center"
+      :class="{
+        'py-5': !loading,
+        'py-4': loading,
+      }">
```
daiagi, line 8: that's a bit odd
can't we stick to one of them?

components/shared/modals/BoxedModal.vue
```diff
@@ -0,0 +1,165 @@
...
+        </div>
+      </div>
+
+      <div
+        ref="slot"
+        class="slot"
+        :class="{
+          slot__loading: loading,
+        }">
+        <slot></slot>
```
daiagi, line 79:
```suggestion
<slot/>
```

components/shared/modals/BoxedModal.vue
daiagi, line 1: i'm not a fan of this name
it isn't a modal, it's a loader wrapper for modal content

the name should reflect that 



plus, should it go inside `@kodadot1/brick` maybe? inside the NeoModal dir?

daiagi (changes requested):
nice work
just a few changes please

components/shared/modals/BoxedModal.vue
hassnian: base is py-5

loading needs to be 4 since the loader is bigger than the text otherwise there is a change of the box size here 

CleanShot 2023-11-20 at 12 30 36

components/shared/modals/BoxedModal.vue
hassnian: fixed

components/shared/modals/BoxedModal.vue
hassnian: fixed

daiagi (approved):
still not a fan of the name tbh
i think it should reflect the fact that it adds loading UI around the slot

hassnian:
> still not a fan of the name tbh
> 
> i think it should reflect the fact that it adds loading UI around the slot
> 
> 
> 
> 

well it does add a prop with loading but you don't have to use it , and can have the title box and the body 

any suggestions

prury:
i would wait for #8136 to be merged before this one, wdyt?

exezbcz:
Is there an estimation of the modal height that is after that? 

- I think in this case we can make the rescalling possible. 
- I did not like that it was before or after the auto teleport loaded. This will be okay (loading modal, once loaded it opens a modal that is slightly larger) 


https://github.com/kodadot/nft-gallery/assets/90852205/69a25f87-f935-403b-8128-0d031a9276b5



Or make the loading modal higher? 


image
because here we are restricting so much of the vertical space imo

hassnian:
> Or make the loading modal higher?

yes, prev max modal height was 50vh now 70vh 

https://github.com/kodadot/nft-gallery/assets/44554284/e0506504-88dc-492e-8004-9625f5d9aa86

exezbcz:
oki oki. we can always make follow-up, tried it now and it's working nicely, thanks! 

ready to be merged? So we can continue to #8163

prury:
blocked as suggested on https://github.com/kodadot/nft-gallery/pull/8124#issuecomment-1819584672

roiLeo:
👀 
Screenshot 2023-11-23 at 16-26-53 remove only if dataset not affected too much by daiagi · Pull Request #8186 · kodadot_nft-gallery

libs/ui/src/components/NeoModal/NeoModalBody.vue
```diff
@@ -0,0 +1,167 @@
...
+        <div class="skeleton-content is-flex">
+          <NeoIcon
+            icon="spinner-third"
+            class="spinner has-text-k-grey mr-6"
+            size="large"
+            spin></NeoIcon>
+
+          <div>
+            <p class="is-capitalized has-text-weight-bold is-size-6">
+              {{ $t('general.doingSomeMagic') }}
```
roiLeo, line 63: no i18n in this package

libs/ui/src/components/NeoModal/NeoModalBody.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <div class="modal-width">
+    <header
+      class="px-6 is-flex is-justify-content-space-between border-bottom is-align-items-center"
+      :class="{
+        'py-5': !loading,
+        'py-4': loading,
```
roiLeo, line 7: why different size between?

libs/ui/src/components/NeoModal/NeoModalBody.vue
```diff
@@ -0,0 +1,167 @@
...
+      }">
+      <NeoSkeleton
+        v-if="loading"
+        rounded
+        width="150"
+        height="35"
+        no-margin
+        border-radius="4rem"
```
roiLeo, line 15: is border-radius prop really needed?

libs/ui/src/components/NeoModal/NeoModalBody.vue
```diff
@@ -0,0 +1,167 @@
...
+const emits = defineEmits(['close'])
+withDefaults(
+  defineProps<{
+    title: string
+    loading?: boolean
+    modalWidth?: string
+    modalMaxHeight?: string
+    paddingTop?: string
+    paddingBottom?: string
+    paddingX?: string
```
roiLeo, line 99: use content-class instead?

roiLeo (commented):
I think there's been a lot of change for a modal, can we do it with less code change? maybe without `NeoModalBody`?

hassnian:
> I think there's been a lot of change for a modal, can we do it with less code change? maybe without `NeoModalBody`?

@roiLeo  I'm not sure what to say, 

you mean this solution has more code than it should ?

could you please elaborate 

thanks

hassnian:
> > > I think there's been a lot of change for a modal, can we do it with less code change? maybe without `NeoModalBody`?
> > 
> > 
> > @roiLeo I'm not sure what to say,
> 
> **maybe no need to add a skeleton to the whole component (title, body). Just add a skeleton to the fee/deposit section** and below. Because loading depends on `const loading = computed(() => !autoteleport.value?.hasBalances)` in this PR. So, we don't need to create a new component for this
> 
> I don't think the title and some components on the body need to wait to load

  I don't mind, makes sense just following the design  cc @exezbcz

> Would you be ready for merge? cc @hassnian

on it

libs/ui/src/components/NeoModal/NeoModalBody.vue
hassnian: this skeleton has rounder borders.

libs/ui/src/components/NeoModal/NeoModalBody.vue
roiLeo: then reduce size of skeleton? it shouldn't affect content spacing/size

libs/ui/src/components/NeoModal/NeoModalBody.vue
hassnian: content-class  is for neomodal not for this section 
CleanShot 2023-11-27 at 18 28 46@2x

which is the bottom section of the modal 
CleanShot 2023-11-27 at 18 29 16@2x

libs/ui/src/components/NeoModal/NeoModalBody.vue
hassnian: I mean i'm ok , but then it wont match the design exactly

libs/ui/src/components/NeoModal/NeoModalBody.vue
roiLeo: This clearly doesn't look right, I find this to specific for a single case.

Maybe we should instead style the content, like how it's done in [Buefy](https://buefy.org/documentation/modal/#component) (modal-card)

hassnian: fair enough

exezbcz:
> I don't mind, makes sense just following the design cc @exezbcz
> 
> >

>I am not really fan of landing a modal that is not loaded. Having a loader and then getting to a modal that is 100% ready is IMO better UX.


Yeah, this was the reason

The thing why the whole modal: rest of the modal apart from the auto teleport does not need to be loaded, so you see that. After teleport loads, it resizes the modal. One step is better for me.  

we can always iterate, but right now it makes sense to me to show this loading modal

libs/ui/src/components/NeoModal/NeoModalBody.vue
hassnian: increasing the height without reducing the top box padding makes the height of the top box change when it stops loading, 

with these two padding, the top box height does not change. when it goes from loading to not loading

roiLeo (commented):
I would move `NeoModalBody` component inside kodadot repo since you add logic & text inside

hassnian:
> I would move `NeoModalBody` component inside kodadot repo since you add logic & text inside

sure, it was prev there I'll move it back

libs/ui/src/components/NeoModal/NeoModalBody.vue
preschian: How about adding `line-height` to the text?

libs/ui/src/components/NeoModal/NeoModalBody.vue
roiLeo: @exezbcz let's keep/unify the same border radius on all skeleton

preschian (approved):

yangwao:
fast tracking 
pay 70 usd

yangwao:
💵 $70 @ 5.16 USD/DOT ~ 13.566 $DOT
🧗 [13QUj3pZyFNfYj4AM336hRdyLQbevj5H3sR4PKmLEXLdwZhh](https://polkadot.subscan.io/account/13QUj3pZyFNfYj4AM336hRdyLQbevj5H3sR4PKmLEXLdwZhh)
🔗 [0x0000f0c756be0b6889c48e848fbd55e6f95f4b053c04c1d919559e98c5171121](https://polkadot.subscan.io/extrinsic/0x0000f0c756be0b6889c48e848fbd55e6f95f4b053c04c1d919559e98c5171121)

PR #8180, opened 2023-11-21
Fix: New disclaimer: Do not send to exchanges on certain networks (AH) #8149

### LINKED ISSUE #8149: Transfer: New disclaimer: Do not send to exchanges on certain networks (AH)
Not the first artist who lost his funds over this (I know about two): Sending their KSM from AH to finance. We need to avoid these losses and do everything to prevent them. Once an artist is trying to send ksm/dot from AH to exchange:

- add a disclaimer: not send to exchanges on certain networks
- or not allow it at all (do we know how to do it?)

ideally, add a info that they need to bridge the funds (and how to do it).

For the future - implement autoteleport somehow to transfers from AH?

AshutoshSingh00001:
Thank you for your contribution to the [KodaDot - One Stop Shop for Polkadot NFTs](https://kodadot.xyz/).

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Closes #8149
- [x] ## Did your issue had any of the "$" label on it?
- [x] My DOT address: [Payout](https://kodadot.xyz/dot/transfer?target=158VTDxZJidzsMcAdh7h828domveRvpmk1SCxwLUEHhcgY4v&usdamount=0&donation=true)

### DIFF
--- a/libs/ui/src/components/NeoAvatar/NeoAvatar.vue
+++ b/libs/ui/src/components/NeoAvatar/NeoAvatar.vue
@@ -8,7 +8,7 @@
     :width="size"
     :height="size"
     :style="customStyle"
-    @error="onError" />
+    @error.once="onError" />
   <img
     v-else
     :src="placeholder"
@@ -38,6 +38,7 @@ const customStyle = computed(() => ({
 const onError = (e: Event) => {
   const target = e.target as HTMLImageElement
   if (target) {
+    target.removeAttribute('srcset')
     target.src = props.placeholder
   }
 }


AshutoshSingh00001:
@JustLuuuu @prury please review

JustLuuuu:
@AshutoshSingh00001  

I wanted to drop my bit here. 
- It's not very clear why people should not send the funds to exchange. 
- We are not offering any solution to them (what should they do if they want to send funds to exchange?)
- The hovering info point has the same text as the main text. 
- I would put a different text under the info circle: Something like if you are sending your funds from Assethub to exchange, your funds may be permanently lost; teleport them before sending. Or something in this style

Screenshot 2023-11-22 at 18 00 43

AshutoshSingh00001:
> @AshutoshSingh00001
> 
> I wanted to drop my bit here.
> 
> * It's not very clear why people should not send the funds to exchange.
> * We are not offering any solution to them (what should they do if they want to send funds to exchange?)
> * The hovering info point has the same text as the main text.
> * I would put a different text under the info circle: Something like if you are sending your funds from Assethub to exchange, your funds may be permanently lost; teleport them before sending. Or something in this style
> 
> Screenshot 2023-11-22 at 18 00 43

thanks for the reply @JustLuuuu 
i'lll be updating my pr according to your needs

JustLuuuu:
AAA text ran away from the info box 😂
Screenshot 2023-11-22 at 18 56 29

prury:
mobile needs work as well: 
image

exezbcz:
Uf, I'm not sure about this. 

Sorry, but I did not drop any instructions, now it's just a random toggle button that does not make sense. 


https://github.com/kodadot/nft-gallery/issues/8149#issuecomment-1817904418

PR #8188, opened 2023-11-22
fix: download image does not work

### LINKED ISSUE #8170: download image doesn't work
@yangwao @prury  @kodadot/internal-dev 

### tl;dr
don't think we can solve this one inside the mobile wallet, it's a permission issue

### the findings

1. I debugged it and found no difference in the running logic between desktop and mobile
2. image **does** get downloaded on regular mobile browser (Firefox, chrome)
3. compare permissions below - Nova is missing required permissions

#### Firefox:

image

#### Nova (Math wallet same)

photo1700560662

Jarsen136:

  ## PR Type

  - [x] Bugfix

  ## Needs QA check

  ## Context

  - [x] Closes #8170

I could not detect if it's in the nova wallet because there is nothing difference on the UA. 
We may also have similar cases on other mobile wallets, so I only check if it's in the mobile device.

  ## Screenshot 📸

  - [ ] My fix has changed UI

  ## Copilot Summary
  
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 08a6bab</samp>

Enhanced the user experience of downloading images on mobile devices by opening them in a new tab instead of saving them locally. Modified the `utils/download.ts` file to use the `isMobileDevice` function.

  
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 08a6bab</samp>

> _Mobile devices_
> _Use `isMobileDevice` now_
> _Open image in spring_

### DIFF
--- a/components/gallery/GalleryItemButton/GalleryItemMoreActionBtn.vue
+++ b/components/gallery/GalleryItemButton/GalleryItemMoreActionBtn.vue
@@ -8,17 +8,13 @@
           icon="ellipsis-vertical"
           :active="active" />
       </template>
-      <component
-        :is="isMobileDevice ? 'a' : 'div'"
+
+      <NeoDropdownItem
         v-if="mimeType?.includes('image') && ipfsImage"
-        v-safe-href="toOriginalContentUrl(ipfsImage)"
-        target="_blank">
-        <NeoDropdownItem
-          data-testid="gallery-item-more-dropdown-download"
-          @click="downloadMedia">
-          Download
-        </NeoDropdownItem>
-      </component>
+        data-testid="gallery-item-more-dropdown-download"
+        @click="downloadMedia">
+        Download
+      </NeoDropdownItem>
 
       <template v-if="accountId === currentOwner">
         <NeoDropdownItem @click="burn">Burn</NeoDropdownItem>
@@ -55,7 +51,15 @@ const props = defineProps<{
 }>()
 
 const downloadMedia = () => {
-  if (!isMobileDevice && props.ipfsImage) {
+  if (props.ipfsImage) {
+    const originalUrl = toOriginalContentUrl(props.ipfsImage)
+    if (isMobileDevice) {
+      toast($i18n.t('toast.downloadOnMobile'))
+      setTimeout(() => {
+        window.open(originalUrl, '_blank')
+      }, 2000)
+      return
+    }
     try {
       downloadImage(toOriginalContentUrl(props.ipfsImage), props.name)
     } catch (error) {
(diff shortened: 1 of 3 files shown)


daiagi:
ok it's better then the previous situation of no response at all (mobile)
but still the image can't be downloaded

prury:
@Jarsen136 it partially works, i mean, it redirects me to the image itself, but not sure how to proceed after that, my mobile doesn't show any option to copy or save, i can just screenshot it

Jarsen136:
https://github.com/kodadot/nft-gallery/assets/31397967/225c090c-7a38-4167-87b3-d9f6c48b2691

prury:
> On ios device, I could long-press on the image and then get the menu. Does it also work on android?
> trim.0ACDDBBC-69D1-4A73-B85B-5AAC57580BD6.MOV

unfortunately, it does not

Jarsen136:
> but still there is no clear way of getting the image into user device

>> On ios device, I could long-press on the image and then get the menu. Does it also work on android?
> unfortunately, it does not

Look like there is a limit on Android nova wallet, so users have to copy the url from the header and then open it on another browser: )

This fix at least provides the url of the original image. And on IOS we could copy url or save the image directly.

daiagi:
>Look like there is a limit on Android nova wallet, so users have to copy the url from the header and then open it on another browser: )

i find this an acceptable solution, but in that case there should be a "copy url" button and some instructions i think

Jarsen136:
> > Look like there is a limit on Android nova wallet, so users have to copy the url from the header and then open it on another browser: )
> 
> i find this an acceptable solution, but in that case there should be a "copy url" button and some instructions i think

I just found that on Android nova, the functionality of `copy link` does not work at all, while on IOS it works.

daiagi (approved):
Still, some explainer text would be welcome

Jarsen136:
> Still, some explainer text would be welcome

✅ I added a toast for it while clicking download.
image

preschian (approved):

yangwao:
thanks!
pay 20 usd

yangwao:
💵 $20 @ 5.13 USD/DOT ~ 3.899 $DOT
🧗 [16SjUbGKSdjCdWTy3NNT3JxbRVGGqD4mwkHpc6BD9U2Rp29Z](https://polkadot.subscan.io/account/16SjUbGKSdjCdWTy3NNT3JxbRVGGqD4mwkHpc6BD9U2Rp29Z)
🔗 [0x65e1123be55e6d5ed56e743e7500aa42294cfbe967d474f4fe2eb1ea7671d7fe](https://polkadot.subscan.io/extrinsic/0x65e1123be55e6d5ed56e743e7500aa42294cfbe967d474f4fe2eb1ea7671d7fe)

PR #8306, opened 2023-11-30
chore: prevent reload on development mode

preschian:
## Context

Hi @kodadot/internal-dev, is it safe to adjust the config with this? I constantly reload every time I start dev mode 🤣

Screenshot 2023-11-30 161429


daiagi:
Yessss, so annoying

PR #8460, opened 2023-12-07
🔧 add new live drop

roiLeo:
- fix landing link
- add to generative carousel
- add to sidebar filter


yangwao:
make sense, thanks

anyway link is spreading "not through landing"

PR #8494, opened 2023-12-08
fix: view collection on drops should not open new window

### LINKED ISSUE #8492: view collection on drops should not open new window
why it is opening in new window when I click on view collection? 

image

Jarsen136:
## PR Type

  - [x] Bugfix

  ## Context

  - [x] Closes #8492

### DIFF
--- a/components/collection/unlockable/UnlockableCollectionInfo.vue
+++ b/components/collection/unlockable/UnlockableCollectionInfo.vue
@@ -23,7 +23,7 @@
           rounded
           :tag="NuxtLink"
           :to="`/${urlPrefix}/collection/${collectionId}`"
-          target="_blank"
+          target="_self"
           icon="arrow-right">
           <span>View Collection</span>
         </NeoButton>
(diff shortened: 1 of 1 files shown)


roiLeo (approved):
`_self` is default you can remove the line

components/collection/unlockable/UnlockableCollectionInfo.vue
```diff
@@ -23,7 +23,7 @@
           rounded
           :tag="NuxtLink"
           :to="`/${urlPrefix}/collection/${collectionId}`"
-          target="_blank"
+          target="_self"
```
Jarsen136, line 26:
```suggestion
```

Jarsen136:
> `_self` is default you can remove the line

✅

exezbcz:
I don't think it's wrong; if wao wants to change it, let's go.

roiLeo:
> I don't think it's wrong; if wao wants to change it, let's go.

Yup I think we should keep open in new tab since if user have already generate some art (but not yet mint) it will loose the generation once click on link.

daiagi (approved):

roiLeo:
tenor

exezbcz:
@roiLeo yeah, there is still cmd click


nevermind, small change

PR #8570, opened 2023-12-14
GraphQL typescript codegen PoC

shashkovdanil:
# It's just PoC so I wanna know your opinion about it

This solution automatically outputs TS types based on the schema and described queries/mutations. On the screenshot you can see code autosuggestion.

Screenshot 2023-12-14 at 13 08 30

If you like this idea, I can customize it for different chains and schemes in the near future (it can be done)

cc @roiLeo @preschian @vikiival

### DIFF
--- a/__generated__/gql.ts
+++ b/__generated__/gql.ts
@@ -1,0 +1,42 @@
+/* eslint-disable */
+import * as types from './graphql';
+import { TypedDocumentNode as DocumentNode } from '@graphql-typed-document-node/core';
+
+/**
+ * Map of all GraphQL operations in the project.
+ *
+ * This map has several performance disadvantages:
+ * 1. It is not tree-shakeable, so it will include all operations in the project.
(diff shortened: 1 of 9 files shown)


preschian:
> typescript codegen

yes, please

> If you like this idea, I can customize it for different chains and schemes in the near future (it can be done)

I also have concerns regarding this, thank you for bringing up the topic. Initially, I thought we needed to merge all the schemas first. Then from there, we could get the code generation results. But I lack experience in this matter.

- https://the-guild.dev/graphql/mesh/docs/getting-started/comparison

vikiival:
> @roiLeo @preschian guys text me please list of schemas (or url to schemas) that we use

Libs/static/indexers.ts

yangwao:
Stale, please @shashkovdanil  reopen if there is intent to finish it.
Most likely worth bounty.

PR #8643, opened 2023-12-17
fix: Different number of items in drops page and drop iteself

### LINKED ISSUE #8639: Different number of items in drops page and drop iteself
I am curious why they are different for same drop item 😄

image

image

Jarsen136:
## PR Type

  - [x] Bugfix
 
  ## Context

  - [x] Closes #8639

image


yangwao:
perfect! works thanks!

Issue #8561, opened 2023-12-13
Search icon sizing on mobile 


exezbcz:
![image](https://github.com/kodadot/nft-gallery/assets/90852205/12d6d4d9-8ca2-478c-b1ce-b3bd33397885)
![image](https://github.com/kodadot/nft-gallery/assets/90852205/2966e1a8-055e-4f7c-b0e7-35823ed8b425)

- lets make it according to the new button sizings
- not sure if this would be fixed by @shashkovdanil migration 

the input is currently set to 40 height - that is correct 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/cec644ea-489e-4272-9dc1-0c6b24f87747)

- lets make the search icon 40x40px as well, centered icon, you can use whatever fontawesome offers 

## close and burger icon 
![image](https://github.com/kodadot/nft-gallery/assets/90852205/892cd102-7aa1-4c50-bc7f-e5ef9bce92cc)
- touchpoint is nice, but the icon is quite small in the container
- ideal would be to use fontawesome regular sharp x-mark set, we will lose the animation no?

PR #8655, opened 2023-12-18
🔧 navbar search square btn

roiLeo:
- ref #8561 

Capture d’écran 2023-12-18 à 9 43 26 AM


preschian (approved):

Jarsen136 (approved):

components/Navbar.vue
```diff
@@ -360,6 +360,11 @@ useEventListener(window, 'resize', handleResize)
 </script>
 
 <style lang="scss" scoped>
+.square-40 {
```
roiLeo, line 363: bad solution

yangwao: why

PR #8588, opened 2023-12-15
💅 sort by volume DESC

roiLeo:
Capture d’écran 2023-12-15 à 10 08 48 AM

- closes #5678

PR #8658, opened 2023-12-18
💅 sort by highestSale DESC

roiLeo:
- continue #8588 

Screenshot 2023-12-18 at 15-02-56 KodaDot - NFT Market Explorer

### DIFF
--- a/components/explore/ExploreSort.vue
+++ b/components/explore/ExploreSort.vue
@@ -79,7 +79,7 @@ const options = computed(() => {
     sortBy = NFT_SQUID_SORT_COLLECTIONS
 
     if (isAssetHub.value) {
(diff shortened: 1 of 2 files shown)


prury:
@roiLeo can you make it clear other sort filters when highest volume is selected? because i don't think it will work with any other filter, or it will? 

persists even after worker PR: 
image

roiLeo:
> @roiLeo can you make it clear other sort filters when highest volume is selected? because i don't think it will work with any other filter, or it will?

will be fixed once indexer release (kodadot/stick#163), don't worry about those collections at the bottom of the page, they will be hidden (legacy unique)

Jarsen136 (approved):

preschian (approved):

PR #8809, opened 2023-12-30
trigger 'buy-now' filter when sorting by price

### LINKED ISSUE #8353: Apply buy now filter when sorting based on price
image
- apply buy now filter when price filter is applied

daiagi:

## PR Type

- [x] Feature

## Needs QA check

## Context

- [x] Closes #8353

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1cAsKZYNRb8dkSCpn4eVkEn6ycNZTGoDo5dGDgB8J1UUWK8)

### DIFF
--- a/components/items/ItemsGrid/useItemsGrid.ts
+++ b/components/items/ItemsGrid/useItemsGrid.ts
@@ -10,7 +10,6 @@ import { nftToListingCartItem } from '@/components/common/shoppingCart/utils'
 
 import { useListingCartStore } from '@/stores/listingCart'
 import { NFT, TokenId } from '@/components/rmrk/service/scheme'
-import { LocationQueryValue } from '#vue-router'
 
 const DEFAULT_RESET_SEARCH_QUERY_PARAMS = [
   'sort',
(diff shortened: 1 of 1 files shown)


exezbcz:
thanks!

components/items/ItemsGrid/useItemsGrid.ts
```diff
@@ -10,6 +10,7 @@ import { nftToListingCartItem } from '@/components/common/shoppingCart/utils'
 
 import { useListingCartStore } from '@/stores/listingCart'
 import { NFT, TokenId } from '@/components/rmrk/service/scheme'
+import { LocationQueryValue } from '#vue-router'
```
Jarsen136, line 13: What's '#vue-router' ? 
How about `import type {
  LocationQueryValue,
} from 'vue-router'
` or just simply using `string` type instead

daiagi: thanks
fixed

components/items/ItemsGrid/useItemsGrid.ts
```diff
@@ -127,6 +127,25 @@ export function useFetchSearch({
       orderBy: getRouteQueryOrDefault(route.query.sort, ['blockNumber_DESC']),
     }
 
+    type SortBy = string | null
```
Jarsen136, line 130: Small stuff. I would move this type of definition outside this function.

Jarsen136 (approved):

components/items/ItemsGrid/useItemsGrid.ts
daiagi: prefer inline it?
 it's sole purpose is to make to fn signature tiny bit easier to read

Jarsen136: No, I do not mean to inline it. I prefer to move the type definition `type SortBy = string | null` to the top of this file ( Line 23). It would be more clear to separate the type and function.

It doesn’t matter, inline is also possible

preschian (approved):

PR #8967, opened 2024-01-12
fix(drop): nft overlap for same user

### LINKED ISSUE #8961: Stacked cards with and without generative art tag
https://kodadot.xyz/ahp/u/1KxzJHoRJHi8EgWVhrGDueeyzg8PyihAvSL3NP6QMcxEve6

What's the difference between left and right

image

hassnian:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] Closes #8961

## Screenshot 📸


components/collection/drop/GenerativePreview.vue
```diff
@@ -60,10 +61,17 @@ const emit = defineEmits(['select'])
...
+const entropyRange = computed<[number, number]>(() => [
+  STEP * props.minted,
+  STEP * (props.minted + 1),
+])
+
 const getHash = (isDefault?: boolean) => {
   const ss58Format = isDefault
     ? chainProperties.value?.ss58Format
-    : getRandomInt(15000)
+    : getRandomIntFromRange(entropyRange.value[0], entropyRange.value[1])
```
Jarsen136, line 73: `chainProperties.value?.ss58Format` How about the default one? It should be also in the available range.

hassnian: CleanShot 2024-01-13 at 08 17 50@2x

first entropyRange `[ 0, 64 ]` includes 0 and 2 which are the default ones for `ahp` and `ahk`

Jarsen136: I would use `entropyRange.value[0]` as the default value to avoid duplicate items minted by a user.

hassnian: but isn't it the same ?

ahp drop -> ss58Format is 0  and the first range is `[ 0, 64 ]` which has 0 , second range will not have 0 anymore, how do you get a duplicate by the same user ?

ahk same here it's 2 part of the first range after that you can't get 2 anymore 

I have no issue with making the change, but I don't see how you can get a duplicate with this code

Jarsen136: > ahp drop -> ss58Format is 0 and the first range is [ 0, 64 ] which has 0 , second range will not have 0 anymore, how do you get a duplicate by the same user ?

There is a case that a user could make the same drops.
If the user does not click the "Generative New" button, he will always get the drop with the default ss58Format value, which is zero. That means he could get the same drops as much as he wants. wdyt ?

hassnian: yep, absolulty 

updated the code https://github.com/kodadot/nft-gallery/pull/8967/commits/e37f3364bd7dda5b7bbd76b50fac0129750c9750

Jarsen136 (approved):

preschian (approved):

vikiival:
@yangwao yolo or after drop?

yangwao:
> olo or after drop?

@vikiival what can go wrong

vikiival:
> @vikiival what can go wrong, not urgent, however important

depends if the calculation is done correctly :)
Because currently calculation works very bad.

vikiival:
Good but it still does not prevent @yangwao to mint 0 / same pieces

```ts
const getHash = (isDefault?: boolean) => {
// rest could be ok
  const initialValue = accountId.value ? encodeAddress(accountId.value, ss58Format) : Date.now() << ss58Format // I do not have better entropy rn 
  return blake2AsHex(initialValue, 256, null, true)
}
```

vikiival:
Guess @prury is OOO, 
@yangwao can you try ?

vikiival:
Yolo hope this wont break anything

Issue #8479, opened 2023-12-08
new Oruga 0.8.0 - breaks build


yangwao:
sad renovate bot stuck 
- https://github.com/kodadot/nft-gallery/pull/8386

`Error:  [vite]: Rollup failed to resolve import "@oruga-ui/oruga-next/src/utils/NoticeMixin" from "/home/runner/work/nft-gallery/nft-gallery/libs/ui/src/components/NeoNotification/NeoNotificationNotice.vue".`

https://github.com/kodadot/nft-gallery/actions/runs/7141921939/job/19450104374?pr=8386#step:4:21

I mean I would wait till 0.8.1 or 0.8.2, yet it's breaking release as stated by @roiLeo

roiLeo:
modal, tooltip & inputitems wiil need some minor rework 
see https://github.com/oruga-ui/oruga/releases/tag/v0.8.0

kodabot:
ASSIGNED - @daiagi 🔒 LOCKED -> Sunday, December 17th 2023, 07:50:42 UTC -> 72 hours

kodabot:
ASSIGNMENT OVER - @daiagi has been unassigned.

yangwao:
any progress on this @daiagi ?

daiagi:
No. I unassigned myself and turned my attention to other issues

yangwao:
hello, @kodadot/internal-dev can someone take it, please?

roiLeo:
> hello, @kodadot/internal-dev can someone take it, please?

IMO not p1, update all other package and dismiss oruga for now it will come later since it could breaks many things

preschian:
1st attempt upgrade oruga https://github.com/kodadot/nft-gallery/pull/8972
the build ✅
the results ❌

yangwao:
Meanwhile, Oruga 0.8.6
https://www.npmjs.com/package/@oruga-ui/oruga-next

PR #8972, opened 2024-01-13
refactor: oruga composition api

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Related #8479


roiLeo:
no advantages in continuing to develop on this branch? is it the new composition api that causing all issue?

preschian:
> no advantages in continuing to develop on this branch? is it the new composition api that causing all issue?

I couldn't make it. This is my 1st attempt to upgrade the Oruga package. On 2nd attempt locally, I tried to use multiple versions of Oruga and wanted to check it one by one. But the result is the same: the component does not render. I think the problem is we can't use `extend:` anymore because Oruga components are now using composition API.

I think the possible way is we need to `"re-implement/re-import"` (idk what the correct terms for this) the component again like the `NeoButton` currently used. But I don't like this approach, to be honest.

preschian:
It seems like the options for us are:
1. Follow the `NeoButton` approach, `"re-implement/re-import"` the component. We need to re-create components using `extends:` or `mixins:` on `@kodadot1/brick`. Because we can't use `extends:` or `mixins:` anymore.
2. Use Nuxt Layers. POC in here: https://github.com/kodadot/nft-gallery/pull/9020.

PR #8973, opened 2024-01-13
style: remove is-size-7

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] remove `is-size-7` css class

### DIFF
--- a/assets/styles/global.scss
+++ b/assets/styles/global.scss
@@ -310,11 +310,6 @@ a.has-text-grey {
     border-right: 1px solid theme('border-color');
   }
 }
-.border {
-  @include ktheme() {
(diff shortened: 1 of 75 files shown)


Jarsen136 (approved):

yangwao:
cc @preschian 
image

cc @prury do please quick check if nothing breaks

prury:
> 
> cc @prury do please quick check if nothing breaks

made a quick check as you said and it seems fine, old is-size-7 is being properly replaced by tailwind class
did not checked them all tho

roiLeo (approved):
Note that text-xs class add a line height

PR #9020, opened 2024-01-16
feat: Nuxt Layers + Oruga

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] POC to test Nuxt Layers + Oruga


preschian:
> wait what? how will you handle style customization?

Since Nuxt Layers is basically a Nuxt App, we can follow their guidelines for customization: https://oruga.io/documentation/themes.html

in [cf4c834](https://github.com/kodadot/nft-gallery/pull/9020/commits/cf4c834f1ad6827b347b673e28e5535a4c2acfad), I override the `$primary` variable. you can check the `Tabs` component on the gallery detail page

roiLeo:
Looks like font does not load as before.. imo this could introduce so many breaking changes.
I get the impression that you're embarking on a development that's far too long and won't really improve the dev experience.
Plus I don't really get the point of using nuxt layer as another internal workspace, I might need to read more ressource about that one.

preschian:
> Looks like font does not load as before.. imo this could introduce so many breaking changes.
> 
> I get the impression that you're embarking on a development that's far too long and won't really improve the dev experience.
> 
> Plus I don't really get the point of using nuxt layer as another internal workspace, I might need to read more ressource about that one.

no worries, this PR only POC. so anyone feel free to replies with agreement/disagrement

for font, it's replaced also to 'monospace' under $primary variable

if the extends/mixins is works, i would go with that approach instead. since currently we can't use that anymore, NeoButton approach is not so straightforward, imo. because need to reimplement the components structure again (pass the props, readd the slot, and some cases. not checked yet about programmatically use case for modal/notification). not only customization

PR #9184, opened 2024-01-30
fix: parentnode

preschian:
**Thank you for your contribution** to the [KodaDot - One Stop Shop for Polkadot NFTs](https://kodadot.xyz).

👇 __ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] Related `https://github.com/vuejs/core/issues/10098`
- [x] Closes #9075
- [x] Closes #9178

#### Did your issue had any of the "$" label on it?

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1xjvRADwdJcnmUCLWerEHR4iGCT5EBTm4D4fzLLg4LcAC9p)

PR #9185, opened 2024-01-30
fix: insertBefore sidebar

preschian:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] Fix insertBefore node in sidebar


preschian:
@roiLeo, in retrospect, I think we need to find a proper workaround for our UI Kit components. If we re-implement the component instead of extending it, It can lead to another bug. ref: https://github.com/kodadot/nft-gallery/pull/9020#issuecomment-1893786868

For example, I change our `NeoModal` to `NeoModalExtend` in this PR. There is some error occurring on `NeoModal`

| NeoModal | NeoModalExtend |
|--------|--------|
| Screenshot 2024-01-30 at 1 20 28 PM | Screenshot 2024-01-30 at 1 21 25 PM | 

How to reproduce:
- visit `/`
- go to the `settings` page from the sidebar
- click `create` on the navbar

roiLeo (approved):

roiLeo:
> @roiLeo, in retrospect, I think we need to find a proper workaround for our UI Kit components. If we re-implement the component instead of extending it, It can lead to another bug. ref: [#9020 (comment)](https://github.com/kodadot/nft-gallery/pull/9020#issuecomment-1893786868)

issue is comming from `NeoModal` implementation or from RampModal? what solution should we provide?

roiLeo:
> From our `NeoModal`. We need to find a way to extend the component for the composition API component. Actually, `NeoModalExtend` way, but for composition API

How big would it be to copy new implementation of composition api from oruga ui? I know some other components could break

preschian:
What does copy mean? Did you mean to re-implement the component like the `NeoModal` way?

The problem was that `NeoModalExtend` didn't work in the newer Oruga version because Oruga migrated their component from Options API to Composition API.

roiLeo:
> What does copy mean? Did you mean to re-implement the component like the `NeoModal` way?

or create a composable to programmatically open/close modals

> The problem was that `NeoModalExtend` didn't work in the newer Oruga version because Oruga migrated their component from Options API to Composition API.

implementation has changed so our code need to adapt to it, new calls seem simpler.

preschian:
Oooh, I mean, I talked about all our general components. How do we support the new version of Oruga?

I like `extends: OCollapse,` way on `NeoCollapse` compared to re-implement the component like `NeoCheckbox`. If we re-implement the component like `NeoCheckbox`, I mean that it can lead to another bug. Like we did in `NeoModal`.

But, `extends: OCollapse` no longer works on the newer Oruga version.

> implementation has changed so our code need to adapt to it, new calls seem simpler.

Oooh, did I miss something? Is there any way to extend the new Oruga component, or did you mean to migrate the component like `NeoCheckbox`?

preschian:
> Ooh, didn't knew about that one, any easy way to extend component with new composition api? 

Yes, we need to find this

> (maybe [defineComponent](https://vuejs.org/guide/essentials/template-refs.html#definecomponent)?)

Will check if it is possible or not with `defineComponent`

preschian:
closing this, already included in this PR https://github.com/kodadot/nft-gallery/pull/9184/commits

PR #9252, opened 2024-02-02
fix: Drop page card hover and default state

### LINKED ISSUE #9248: Drop page card hover and default state
image

image
from figma: 

- there should be one stroke between the bottom section and the img
- stroke default is grey, active black 
- no change of font color 

link: https://www.figma.com/file/JXNtryaNYhY0XJNArzvPQx/Handoff?type=design&node-id=394-6126&mode=design&t=TUfbyfRyvQIpdXlD-11

Jarsen136:

  ## PR Type

  - [x] Bugfix

  ## Needs Design check

  - @exezbcz please review

  ## Context

  - [x] Closes #9248

  ## Screenshot 📸

  - [x] My fix has changed UI

image

image

  ## Copilot Summary
  copilot:summary

  copilot:poem


preschian (approved):

exezbcz:
default state correct, have a please look at the active once again 

image


text color should not change plus only the outer stroke should change to black

Jarsen136:
will check

Jarsen136:
> default state correct, have a please look at the active once again
> 
> image
> 
> text color should not change plus only the outer stroke should change to black

✅ Fixed

components/drops/DropCard.vue
```diff
@@ -130,10 +131,7 @@ onMounted(async () => {
 .drop-card {
   &:hover {
     .drop-card-banner {
```
roiLeo, line 133: this does not look good as `&-banner` is defined underneath, I would add `&:hover` inside it or use class group hover

otherwise ok

roiLeo (approved):

components/drops/DropCard.vue
Jarsen136: There is detail that when hovering on the whole card `drop-card`,  the style should change. That's why I could not add `&:hover` inside only the banner `drop-card-banner`.  wdyt?

roiLeo: I would consider refactoring to something cleaner since I find it kinda hard to read

Jarsen136: lmk if you have any idea to make it with a better solution.

exezbcz:
this card will be replaced after: 
- #9277

we can merge this one IMO

PR #9577, opened 2024-02-28
⚠️ do not merge/arrowsquid ahp

vikiival:
- **:wrench: arrow stick AHP**

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

## Needs QA check

## Context

- [ ] Closes #<issue_number>

## Screenshot 📸

- [ ] My fix has changed UI

### DIFF
--- a/utils/constants.ts
+++ b/utils/constants.ts
@@ -52,7 +52,7 @@ export const URLS = {
     rubick: 'https://squid.subsquid.io/rubick/graphql',
     marck: 'https://ksm.gql.api.kodadot.xyz/',
     stick: 'https://ahk.gql.api.kodadot.xyz/',
-    speck: 'https://ahp.gql.api.kodadot.xyz/',
+    speck: 'https://squid.subsquid.io/speck/v/v11/graphql/',
     polkassembly: 'https://squid.subsquid.io/polkadot-polkassembly/graphql',
     gsPolkdot: 'https://squid.subsquid.io/gs-main-polkadot/graphql',
     gsKusama: 'https://squid.subsquid.io/gs-main-kusama/graphql',


yangwao:
@vikiival, what are you waiting for here?

vikiival:
- https://github.com/kodadot/stick/issues/220#issuecomment-1989040942

Made another release

vikiival:
If it is OK let's close it then

vikiival:
Ah I forgot to enable ws, thanks for the test

vikiival:
issue was in the trailing slash [e73dc2a](https://github.com/kodadot/nft-gallery/pull/9577/commits/e73dc2afec3ff175248e5a70fbcb5f40768abce8)

yangwao:
Needs to be tested and then can be promoted to production?

vikiival:
> Needs to be tested and then can be promoted to production? 

Exactly, once we close this PR i will promote arrowsquid to prod

prury:
@vikiival i'm getting different query results: 

image

vikiival:
> @vikiival i'm getting different query results:

I have missed the PR from @daiagi 

- https://github.com/kodadot/stick/pull/251

Should be fixed in ~30 min

vikiival:
Probably fixed ? 

kudos to @daiagi 
Screenshot 2024-03-22 at 13 52 17

yangwao:
putting it live?

prury:
> Probably fixed ?
> 
> kudos to @daiagi Screenshot 2024-03-22 at 13 52 17

yup, this is fixed

only small difference in query results now (21) items

vikiival:
> only small difference in query results now (21) items on explore -> items

I am not sure what you mean. Can you provide screenshots please?

vikiival:
> Same as this, but with smaller difference

I suppect there is a huge hidden bug in the token entity (cc @daiagi )

daiagi:
Should be some sql issue in the total token entity resolver, nothing major (in terms of effort to fix) 
Can assume that previously number is the correct one as I checked it well when I first wrote it. And the previous query was simple

daiagi:
Something completely different is going on here and it's not so easy to explain over gh post

1. it isn't related to tokenEntity at all, and token entity is reporting correct numbers (compare PR and canary with "the break" search query)

2. it's about metadata mismatch between canary and pr env. no idea why. maybe because the indexer is stuck?


let's walk through it, we will focus on Motifs collection

1. in explore page, search for "motifs"
2. notice that `count@pr = count@canary +4`
3. scroll to bottom and find a stacked nft with countof 8@canary and count of 4@PR. this account for all the mismatch
4. go to collection page (of motifs), collection page does not use TokenEntity at all, notice that total count is same in both envs
5. manually count the amount of stacked nfts, you will find that the count on the "stacked" card was individually correct in each env
6. think to yourself wtf?
7. find out that nft `108-88`, shows the default drop picture on canary (stacked) and **NOT** the default picture on PR env (not stacked)


image



**tl;dr**


some nfts have default image on canary and not default on PR env, which leads to a different count in explore page due to "stacked" cards @vikiival  ant thoughts? my only idea is due to stuck indexer

also, canary env does not impl yet the "tokenEntity" id change, that fixed "tokenEntity" bunching together nfts that should not be bunched together, that also leads to count mismatch

vikiival:
> also, canary env does not impl yet the "tokenEntity" id change, that fixed "tokenEntity" bunching together nfts that should not be bunched together, that also leads to count mismatc

Except this line I was able to understand the issue you have described above ^-^, thanks for the explanation.

probably this would be a correct thing to do 
- https://github.com/kodadot/stick/issues/255

vikiival:
Side effects are my favs 

Screenshot 2024-03-29 at 14 07 01

Issue #8767, opened 2023-12-27
use css helpers to toogle visiblity


roiLeo:
maybe use css helpers to toogle visiblity?

_Originally posted by @daiagi in https://github.com/kodadot/nft-gallery/pull/8758#discussion_r1436785215_

replace `isMobile` and `!isMobile` with visibility helper when we can

- https://tailwindcss.com/docs/visibility#breakpoints-and-media-queries

better to use css than js

pbkompasz:
:wave:

kodabot:
ASSIGNED - @pbkompasz 🔒 LOCKED -> Monday, February 19th 2024, 03:59:19 UTC -> 36 hours

yangwao:
Some work was done in 
- https://github.com/kodadot/nft-gallery/pull/9480

Free to take

kodabot:
ASSIGNMENT EXPIRED - @pbkompasz has been unassigned.

Haviles04:
I think it might be better to use display properties, unless the directive is a v-show.

v-if =  completely removed from dom (display properties)
v-show = still in dom but invisible (visibility properties)

PR #9655, opened 2024-03-07
Fix/multiple-reloads

### LINKED ISSUE #9648: Monotif Test drop generating various previews on refresh
https://canary.kodadot.xyz/ahk/drops/monotifs

https://github.com/kodadot/nft-gallery/assets/36627808/8386401f-86a2-489d-9630-61155125f3a6

daiagi:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Closes #9648

## Links
[monotifs](https://deploy-preview-9655--koda-canary.netlify.app/ahk/drops/monotifs)
[alteration](https://deploy-preview-9655--koda-canary.netlify.app/ahk/drops/alteration)

## Notes
1. Usually it is better to use css to toggle visibilty, but not in this case. `GenerativePreview` is emitting messages, having 2 instances in the page is causing unexpcted behviours
2. yes, it is ugly to use `setTimeout` but in makes the first variation of the art to load consistently, i guess the code in the iframe needs a moment to load

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1cAsKZYNRb8dkSCpn4eVkEn6ycNZTGoDo5dGDgB8J1UUWK8)

## Screenshot 📸

- [ ] My fix has changed UI

### DIFF
--- a/components/collection/drop/GenerativeLayout.vue
+++ b/components/collection/drop/GenerativeLayout.vue
@@ -27,7 +27,7 @@
           <hr class="hidden md:block mt-7 mb-0" />
 
           <CollectionDropGenerativePreview
-            class="md:hidden mt-7"
+            v-if="width < mdBreakpoint"
             :minted="userMintedCount"
             :drop="drop"
             :collection-id="collectionId"
@@ -57,7 +57,9 @@
           <CollectionUnlockableTag :collection-id="collectionId" />
         </div>
 
-        <div class="column hidden md:flex justify-end mt-[-213px]">
+        <div
+          v-if="width >= mdBreakpoint"
+          class="column md:flex justify-end mt-[-213px]">
           <CollectionDropGenerativePreview
             :minted="userMintedCount"
             :drop="drop"
@@ -102,6 +104,7 @@ import type {
 import { useCollectionMinimal } from '@/components/collection/utils/useCollectionDetails'
 import useCursorDropEvents from '@/composables/party/useCursorDropEvents'
 import { DropEventType } from '@/composables/party/types'
+import { useWindowSize } from '@vueuse/core'
 
 const props = withDefaults(
   defineProps<{
@@ -126,6 +129,9 @@ const props = withDefaults(
     holderOfCollection: undefined,
   },
 )
+
+const { width } = useWindowSize()
+const mdBreakpoint = 768
 
 const { emitEvent, completeLastEvent } = useCursorDropEvents(props.drop.alias)
 const { collection: collectionInfo } = useCollectionMinimal({
--- a/components/collection/drop/GenerativePreview.vue
+++ b/components/collection/drop/GenerativePreview.vue
@@ -180,15 +180,11 @@ watch(imageDataLoaded, () => {
   }
 })
 
-watch(
-  accountId,
-  () => {
-    generateNft()
-  },
-  {
-    immediate: true,
-  },
-)
+watch(accountId, generateNft)
+
+onMounted(() => {
+  setTimeout(generateNft, 1000)
+})
 
 watchDebounced(
   [imageDataPayload],
--- a/components/collection/drop/GenerativePreview.vue
+++ b/components/collection/drop/GenerativePreview.vue
@@ -183,7 +183,7 @@ watch(imageDataLoaded, () => {
 watch(accountId, generateNft)
 
 onMounted(() => {
-  setTimeout(generateNft, 1000)
+  setTimeout(generateNft, 100)
 })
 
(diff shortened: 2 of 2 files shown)


components/collection/drop/GenerativeLayout.vue
```diff
@@ -127,6 +130,9 @@ const props = withDefaults(
   },
 )
 
+const { width } = useWindowSize()
```
preschian, line 133: why do we introduce this again? unrelated changes?

ref: https://github.com/kodadot/nft-gallery/issues/8767

components/collection/drop/GenerativePreview.vue
```diff
@@ -180,15 +180,11 @@ watch(imageDataLoaded, () => {
...
-  () => {
-    generateNft()
-  },
-  {
-    immediate: true,
-  },
-)
+watch(accountId, generateNft)
+
+onMounted(() => {
```
preschian, line 185: do these changes fix the issue? it was reloaded because of this line right? https://github.com/kodadot/nft-gallery/blob/684296e5550708088f24ef1863f8c26b0169ce73/components/collection/drop/GenerativePreview.vue#L192-L194

components/collection/drop/GenerativeLayout.vue
daiagi: not unrelated at all

see my notes above in PR description

components/collection/drop/GenerativePreview.vue
daiagi: >do these changes fix the issue?

yes, that's why i made them

please, try for yourself 

>it was reloaded because of this line right

yes, but simple removal of it introduces the problem you solved by adding it

addition of the timeout causes your watcher to never trigger as far as i could tell while playing with it

if you wanna YOLO we can remove the watcher

preschian: I think it was fixed with your changes on `components/collection/drop/GenerativeLayout.vue`, right? so, no need this changes?

preschian: > addition of the timeout causes your watcher to never trigger as far as i could tell while playing with it

I couldn't see any difference in this. It seems like both still trigger the watcher. Or is there a code example I can test?

daiagi: 1. without it, first image shown is always same (not a unique variation)
image

2.  with `immediate: true`, this happens often (not always)
```js
watch(accountId, generateNft, { immediate: true })

// onMounted(() => {
//   setTimeout(generateNft, 1000)
// })
```

image


**3. with setTimeout, first image is unique and loads well**

preschian: Oh, maybe it's way harder to reproduce? It already looks good on my local if I revert this file. But, let's proceed with this anyway

daiagi: i get why you don't like the `setTimeout`
I'm not too big on it either

but it produces better end result in this case

preschian (approved):
> Usually it is better to use css to toggle visibilty, but not in this case. GenerativePreview is emitting messages, having 2 instances in the page is causing unexpcted behviours

nice finding, this one quite important issue tho 👍🏻

components/collection/drop/GenerativePreview.vue
preschian: > i get why you don't like the setTimeout
> without it, first image shown is always same (not a unique variation)

nah, I'm good as long it can fix the issue. I think it's because you already fixed the important issue with the race condition on emit, seems like it's quite harder to reproduce (good thing). It looks good on my local, even though I always get a unique variant if I revert this file. 🤣

> but it produces better end result in this case

yes, I don't want to take a risk. Let's continue with it.

Jarsen136 (approved):

yangwao:
Thanks!
pay 30 usd

yangwao:
💵 $30 @ 10.31 USD/DOT ~ 2.91 $DOT
🧗 [1cAsKZYNRb8dkSCpn4eVkEn6ycNZTGoDo5dGDgB8J1UUWK8](https://polkadot.subscan.io/account/1cAsKZYNRb8dkSCpn4eVkEn6ycNZTGoDo5dGDgB8J1UUWK8)
🔗 [0x95f8d068e744eee9109cb496b284800a58e37bef7a84803d751d731b031353fe](https://polkadot.subscan.io/extrinsic/0x95f8d068e744eee9109cb496b284800a58e37bef7a84803d751d731b031353fe)

PR #9749, opened 2024-03-14
AutoTeleportActionButton: refactoring dependencies.

### LINKED ISSUE #9695: refactor autoteleport
## Current Situation

1. All of AutoTeleport state and function is bounded to `AutoTeleportActionButton`
2.  the chain of calls to composables starting from `useAutoTeleport` is messy and hard to read/ reason
3. for example fees is being drilled through 3 composables to affect `isReady` then is then being drilled up 3 composables, instead there should be a simple function where fees changes `autoTeleportStore.isReady` 

## What i would like to see
1. identify refs and computed refs that belong in store, move them to store
2. function over these state items can be in simple composables / util functions / store actions 
3. composables should have no args or only very simple args (`useIsChain` as an example for simple arg)
4. component can then use multiple composables or use store directly to get whatever it needs

Temidayo32:
`isReady` and the chain of computed values it depends on is implemented in `useAutoTeleportTransitionDetails`. This refactor makes the code more readable by directly importing into `AutoTeleportActionButton` its dependencies from their source file.

This refactor didn't follow as stated in #9695. This is because of the explanation stated above as regards `isReady`, which also applies same to some other dependencies.  Secondly, it's not very clear if `isReady` can be seperated into its own composable as it depends on the long chain of computed values before it in the `useAutoTeleportTransitionDetails` file.

I would appreciate review on this PR, if there is a misunderstanding as regards this issue.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Closes #9695

- [ ] I've merged recent default branch -- **main** and I've no conflicts

#### Optional

- [ ] I've found edge cases

- [x] Fill up your DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=14rs1iTduAn9YrqtDwsTZxK9YDbkzDekBstqxc78M6An87fX&usdamount=0)

### DIFF
--- a/components/common/autoTeleport/AutoTeleportActionButton.vue
+++ b/components/common/autoTeleport/AutoTeleportActionButton.vue
@@ -78,6 +78,7 @@
 import { NeoButton, NeoSwitch } from '@kodadot1/brick'
 import AutoTeleportWelcomeModal from './AutoTeleportWelcomeModal.vue'
 import useAutoTeleport from '@/composables/autoTeleport/useAutoTeleport'
+import useAutoTeleportTransitionDetails from '@/composables/autoTeleport/useAutoTeleportTransitionDetails'
 import type {
   AutoTeleportAction,
   AutoTeleportFeeParams,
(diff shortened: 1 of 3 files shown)


daiagi:
>it's not very clear if isReady can be seperated into its own composable as it depends on the long chain of computed values before it in the useAutoTeleportTransitionDetails file


this long chain is exactly the issue to be solved

as i said, this is not an easy issue and @hassnian is probably the best person for the task

daiagi (approved):
looks like a good first step to me but far from closing the issue though

Temidayo32:
> this long chain is exactly the issue to be solved

Oh, thanks for the clarification.

hassnian:
@daiagi I'm not sure about this, even though code wise this a step in the right direction

**if this is merged like it is now** , since `useAutoTeleportTransitionDetails` is now being called in 3 places 

- AutoTeleportActionButton.vue
- useAutoTeleport.ts
- useAutoTeleportTransition.ts

this is unnecessarily re-fetching data that we already have, like balances and tx fees(both action and telpeort) 

#### opening the listing cart 

CleanShot 2024-03-15 at 14 19 36@2x

#### interacting with the listing cart , refetches actions tx fees 3 times , isn't this also a step backwards?

CleanShot 2024-03-15 at 14 23 52

wdyt?

prury:
can be reopen if internal dev have any response regarding hassnian comment

PR #9838, opened 2024-03-21
Revert "release update"

yangwao:
Reverts kodadot/nft-gallery#9837

cc @prury 

telegram-cloud-photo-size-4-6044302085933089001-x

image


prury:
> Reverts #9837
> 
> cc @prury
> 

checked some PRs and don't know where its coming from
but i can only test on other test drops, since this specific drop doesn't load on my computer as stated before

Issue #9596, opened 2024-02-29
Disable teleport and autoteleport on Polkadot <-> Polkadot Assethub


prury:
The same bug that happened on Kusama in the last month(XCM related) is also happening on Polkadot where users lose funds when trying to teleport an amount near ED. 

let's disable teleport and autoteleport until there's a fix from parity

iMac7:
:wave:

kodabot:
ASSIGNED - @iMac7 🔒 LOCKED -> Friday, March 1st 2024, 17:38:52 UTC -> 24 hours

yangwao:
+50% of ED wouldn't help?

iMac7:
> +50% of ED wouldn't help?

@vikiival 👀

vikiival:
I think we would need to keep 2 * ED at least 

So MAX would be = totalBalance - 2 * ED

iMac7:
> I think we would need to keep 2 * ED at least
> 
> So MAX would be = totalBalance - 2 * ED

How about for autoteleport, check if the amount to be teleported in > 2 * ED ?

prury:
@kodadot/internal issue open for taking

kodabot:
ASSIGNMENT EXPIRED - @iMac7 has been unassigned.

dudo50:
We have checked delivery fees and found following information:

Polkadot delivery fee: 0.047DOT
PolkadotAssetHub delivery fee: 0.036DOT
KusamaAssetHub delivery fee:  0.0015KSM

We are unable to confirm if these values are 100% correct so needs further testing.

exezbcz:
so, another user reported he lost funds, @prury will confirm if its correct. 

Probably let's raise the ed as teleporting from Polkadot to ahp is quite crucial for us


**ad autoteleport**
from what I remember, there was a feature that basically enabled the transfer of the whole amount to the chain the purchase is on to avoid funds loss. Does that cause the fund loss? 


Are there any other solutions?

prury:
> so, another user reported he lost funds, @prury will confirm if its correct.

yup, correct:
https://polkadot.subscan.io/extrinsic/19872686-2
https://polkadot.subscan.io/extrinsic/19881888-3

dudo50:
@exezbcz 
> Probably let's raise the ed as teleporting from Polkadot to ahp is quite crucial for us

[...]

PR #9928, opened 2024-03-29
chore: disable auto teleport

preschian:

## PR Type

- [x] Bugfix

## Context

- [x] Related: https://github.com/kodadot/nft-gallery/issues/9596

I'm not familiar yet with auto-teleport stuff. Let's turn it off until the proper fix.

### DIFF
--- a/components/common/autoTeleport/AutoTeleportActionButton.vue
+++ b/components/common/autoTeleport/AutoTeleportActionButton.vue
@@ -129,6 +129,7 @@ const props = withDefaults(
 const preferencesStore = usePreferencesStore()
 const { $i18n } = useNuxtApp()
 const { chainSymbol, name } = useChain()
+const { flagAutoTeleport } = useFlags()
 
 const amount = ref()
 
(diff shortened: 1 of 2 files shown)


components/common/autoTeleport/AutoTeleportActionButton.vue
```diff
@@ -307,6 +309,10 @@ watch(allowAutoTeleport, (allow) => {
   if (allow) {
     autoTeleport.value = true
   }
+
+  if (!flagAutoTeleport.value) {
+    autoTeleport.value = false
+  }
```
hassnian, line 315: since 

````
const autoTeleport = ref(false)
````


```suggestion
  if (!flagAutoTeleport.value) {
    return
  }

  if (allow) {
    autoTeleport.value = true
  }
```

yangwao:
That's not a good solution as the majority of new people got DOT on the relay chain, then for mints people will.have hard time to mint.

Rather implement @dudo50 suggested solution 

> The only option for now is for the user to have at least 1.5DOT on Polkadot and only transfer sum they have above 1.5 DOT

preschian:
> That's not a good solution as the majority of new people got DOT on the relay chain, then for mints people will.have hard time to mint.
> 
> Rather implement @dudo50 suggested solution 
> 
> > The only option for now is for the user to have at least 1.5DOT on Polkadot and only transfer sum they have above 1.5 DOT 


maybe @Jarsen136 @hassnian could help with that?

PR #9998, opened 2024-04-05
fix: Listed filter being added after route change

### LINKED ISSUE #9995: Listed filter being added after route change
not sure how, but i was able to get this result twice when testing PR: (buy now filter appearing on collections)

still trying to understand how to reproduce it

image

how to reproduce: 

https://github.com/kodadot/nft-gallery/assets/36627808/706fa301-ca2a-4a84-9c27-3f4c4181da69

_Originally posted by @prury in https://github.com/kodadot/nft-gallery/issues/9959#issuecomment-2035368514_

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Closes #9995

## Screenshot 📸

- [ ] My fix has changed UI


hassnian (commented):
did some debugging and found out that 

search.vue

CleanShot 2024-04-05 at 15 21 30@2x

it's adding the listed true , we should have maybe a global middleware that parsed the query params based on the route that we are, 

listed query param does not make sense inside the collections page

also this runs twice since we have 2 `search.vue` components.

additionally this would improve the page loading speed , I'd guess , 


or 


find a quick fix inside the `search.vue` component  , which is adding the query params regardless of the page that we are 

wdyt?

Jarsen136:
> did some debugging and found out that
> 
> search.vue
> 
> CleanShot 2024-04-05 at 15 21 30@2x
> 
> it's adding the listed true , we should have maybe a global middleware that parsed the query params based on the route that we are,
> 
> listed query param does not make sense inside the collections page

Thanks for debugging more!

Making a global middleware would be too much to fix this edge case bug. There are no such kinds of params issues on other pages so far.

> 
> also this runs twice since we have 2 `search.vue` components.
> 
> additionally this would improve the page loading speed , I'd guess ,
> 
It could be a separate performance issue. Only keeping one search instance should improve the perf.


> or
> 
> find a quick fix inside the `search.vue` component , which is adding the query params regardless of the page that we are
> 
> wdyt?


If I fix it inside the search component, I have to add some `if-else` to check the current route. 
Also, I can even reproduce the bug with the following link. It's not the scope inside the search component.
`https://canary.kodadot.xyz/ahp/explore/collectibles?listed=true`

I guess the current solution is easier and simpler to fix this edge issue.

daiagi (approved):
i'm not  fan of the code style choice

1. I would explicitly check for undefined in the `if` condition
2. or skip the if entirely
```js
return navigateTo({ path: to.path, query: {...to.query, listed: undefined })
```

Jarsen136:
> i'm not fan of the code style choice
> 
> 1. I would explicitly check for undefined in the `if` condition

Updated, It makes sense.


> 2. or skip the if entirely

It would cause unnecessary navigate

Jarsen136:
@prury Let's test if the bug is fixed here. Then we can merge it as there is already an approve. I guess there is not much value in discussing more for this small bug.

hassnian (approved):

PR #10067, opened 2024-04-13
fix: Not possible to play with variations on drops that are minted out

### LINKED ISSUE #10050: Not possible to play with variations on drops that are minted out
Its not possible to play with generate button once the collection is minted out? It's generating the same picture. I think it should still work so people can play with the code. Right?

https://github.com/kodadot/nft-gallery/assets/104380776/89925e5c-a3ec-4ecf-bcd0-ad405798df2b

hassnian:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

`mintedAmountForCurrentUser` value was wrong

- [x] Closes #10050

## Needs QA check

## Screenshot 📸

CleanShot 2024-04-13 at 12 21 22


composables/drop/useGenerativeDropMint.ts
```diff
@@ -26,6 +26,7 @@ export type UnlockableCollectionById = {
 export function useCollectionEntity(collectionId?: string) {
   const { drop } = useDrop()
   const { client } = usePrefix()
+  const { accountId } = useAuth()
```
vikiival, line 29: Why do you need account? 

What if account is undefined / user not logged in

hassnian: > Why do you need account?

from this query we get the data that we later use  for `mintedAmountForCurrentUser` and that's why need the accountId


CleanShot 2024-04-15 at 11 59 17@2x


CleanShot 2024-04-15 at 11 58 42@2x


> What if account is undefined / user not logged in

accountId will be '' and 0 will be returned. 

CleanShot 2024-04-15 at 12 01 00@2x

vikiival (approved):

daiagi (approved):

PR #10270, opened 2024-05-10
fix: Kodadot weekly round up on Beehiiv is not there anymore

springshine0925:


hassnian:
@springshine0925 pls use [this pr template](https://github.com/kodadot/nft-gallery/blob/main/.github/PULL_REQUEST_TEMPLATE/PULL_REQUEST_TEMPLATE.md)

daiagi:
1.  I don't think the point was to remove the link, I'd guess the point was to fix beehiiv page
2.  no reason why `pnpm-lock` should change i this PR

springshine0925:
Thank you for your message.
To fix this issue, first of all, we can check their Beehive.com account.
As you can check, this is only link tag and so I have checked that
subdomain but that isn't work.
How do you think about that?
Let me know your advice asap. Thanks.

PR #10306, opened 2024-05-15
feat: Control return to confirm action

### LINKED ISSUE #10213: Control return to confirm action
We should introduce something like that for modals 

apply this action for all confirmation modals

command+return/control+return

Jarsen136:

## PR Type

- [x] Feature

## Needs Design check

- @exezbcz please review

## Needs QA check

## Context

- [x] Closes #10213

## Screenshot 📸


hassnian (commented):
bug or a feature :)

probalby we should only allow that shortcut if the transaction has not started

https://github.com/kodadot/nft-gallery/assets/44554284/98210909-6f95-44cd-839b-74b6ebccb518

would [useMagicKeys](https://vueuse.org/core/useMagicKeys/) [combinations](https://vueuse.org/core/useMagicKeys/#combinations) be a cleaner solution ?

prury:
> bug or a feature :)
> 
> probably we should only allow that shortcut if the transaction has not started
> CleanShot.2024-05-16.at.17.18.47.mp4
> 
> can [useMagicKeys](https://vueuse.org/core/useMagicKeys/) [combinations](https://vueuse.org/core/useMagicKeys/#combinations) give you a cleaner solution?

thank you very much for checking it @hassnian

Jarsen136:
> bug or a feature :)
> 
> probably we should only allow that shortcut if the transaction has not started
> 
> can [useMagicKeys](https://vueuse.org/core/useMagicKeys/) [combinations](https://vueuse.org/core/useMagicKeys/#combinations) give you a cleaner solution?

Thank you! I have updated it.

prury:
What should be the correct behavior? working only for confirmation buttons on modals or also for trying again when a transaction is cancelled/fails? 

Transfer: doesn't work on confirmation button prob because it uses a different modal
Teleport bridge: works on try again
Drop: works on modal and on try again
Create NFT/Collection: won't proceed to signing if you hit it once, i think it's a focus issue, doesn't work on try again
Buy Now: works for signing but doesn't for try again
Listing: works for signing but doesn't for try again

Jarsen136:
> What should be the correct behavior? working only for confirmation buttons on modals or also for trying again when a transaction is cancelled/fails?

I prefer it only works for confirmation buttons on modal.
> 

> Transfer: doesn't work on confirmation button prob because it uses a different modal 

Yup, it's a separate modal. However, I have also made the transfer confirm modal work with the shortcut. Now it should work as the others.


> Teleport bridge: works on try again 
Drop: works on modal and on try again 
Create NFT/Collection: won't proceed to signing if you hit it once, i think it's a focus issue, doesn't work on try again 
Buy Now: works for signing but doesn't for try again 
Listing: works for signing but doesn't for try again

Only applying to the confirm button should be enough.
I can't manage to go to the page with "try again" button after I cancel the transaction request. It keeps loading all the time.
I would leave it to another follow-up issue if anyone else could reproduce it and fix it.

Kapture 2024-05-20 at 01 03 48

hassnian:
modal animation is gone

this pr

https://github.com/kodadot/nft-gallery/assets/44554284/ae5ba16f-7caf-46b3-92eb-530a476c1787

canary


https://github.com/kodadot/nft-gallery/assets/44554284/19daf17d-7c40-4b31-9da0-648cef8340de

libs/ui/src/components/NeoModal/NeoModal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <teleport to="body" :disabled="!appendToBody">
+  <teleport v-if="isModalActive" to="body" :disabled="!appendToBody">
```
hassnian, line 2: can we do a local solution for each modal

composables/useKeyboardKeys.ts
```diff
@@ -0,0 +1,11 @@
+import { useMagicKeys } from '@vueuse/core'
+
+export function useKeyboardKeys({ onPressControlEnter }) {
```
hassnian, line 3: maybe `enabled` param

libs/ui/src/components/NeoModal/NeoModal.vue
Jarsen136: Why apply it for each modal? It's kind of a bug from `OModal` that the DOM of modal will not be removed after closing the modal.

components/common/autoTeleport/AutoTeleportActionButton.vue
```diff
@@ -304,6 +303,10 @@ const handleAutoTeleportModalClose = (completed: boolean) => {
   emit('modal:close', completed)
 }
 
+useKeyboardKeys({
+  onPressControlEnter: handleSubmit,
+})
```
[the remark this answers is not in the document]
Jarsen136: It's not needed, because the keyboard event would be listened to only when the action button component is mounted.

components/transfer/TransferConfirmModal.vue
```diff
@@ -152,6 +152,11 @@ const confirmTransfer = () => {
   closeModal()
   emit('confirm')
 }
+
+useKeyboardKeys({
+  onPressControlEnter: confirmTransfer,
+})
```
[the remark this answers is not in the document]
Jarsen136: the same reason as the above comment

libs/ui/src/components/NeoModal/NeoModal.vue
hassnian: pls [check](https://github.com/kodadot/nft-gallery/pull/10306#issuecomment-2120333504) 

@daiagi what do you think.

components/common/autoTeleport/AutoTeleportActionButton.vue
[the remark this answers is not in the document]
hassnian: I was suggesting this if we removed the teleport `if` 
CleanShot 2024-05-20 at 17 45 27@2x

prury:
rechecking PR:

on transfer, modal is being shown behind the other modal: 

[modal being shown behind.webm](https://github.com/kodadot/nft-gallery/assets/36627808/fa78ad90-cea6-44cc-9b12-88add113d908)

- still working for try again on teleport
- create nft/collection: modal opens behind 

buy now and listing are good

Jarsen136:
Thanks for all the comments above. I will close this PR and leave it to the other one who has a better solution.

PR #10336, opened 2024-05-23
refactor: cleanup free drops component

preschian:
**Thank you for your contribution** to the [Koda - Generative Art Marketplace](https://kodadot.xyz).

👇 __ Let's make a quick check before the contribution.

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Needs QA check

- @kodadot/qa-guild please review

## Context

- [x] Gradual refactor. Related: https://github.com/kodadot/nft-gallery/issues/9575

#### Did your issue had any of the "$" label on it?

- [ ] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=<My_Polkadot_Address_check_https://github.com/kodadot/nft-gallery/blob/main/REWARDS.md#creating-your-dot-address>)

## Screenshot 📸

- [ ] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

PR #10148, opened 2024-04-25
refactor: adjust hash for metadata

preschian:
**Thank you for your contribution** to the [Koda - Generative Art Marketplace](https://kodadot.xyz).

👇 __ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix

## Context

- [x] Related: https://github.com/kodadot/ops-internal/issues/1659
- [x] Need workers deployment: https://github.com/kodadot/private-workers/pull/165

#### Did your issue had any of the "$" label on it?

- [x] My DOT address: [Payout](https://canary.kodadot.xyz/dot/transfer/?target=1xjvRADwdJcnmUCLWerEHR4iGCT5EBTm4D4fzLLg4LcAC9p)

PR #10344, opened 2024-05-24
refactor: cleanup preview items on drops page

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

(please remove design and QA checks below if not needed)

## Needs Design check

- @exezbcz please review

## Needs QA check

## Context

- [x] Related: https://github.com/kodadot/nft-gallery/pull/10336
- [x] Related: https://github.com/kodadot/nft-gallery/pull/10148. The user can't see the preview drops before minting it, ref: https://github.com/kodadot/nft-gallery/pull/10148#issuecomment-2126914884.
- [x] Closes: https://github.com/kodadot/nft-gallery/issues/10352

## Screenshot 📸


exezbcz:
linking: 
- https://github.com/kodadot/nft-gallery/issues/10352

components/collection/drop/modal/PaidMint.vue
```diff
@@ -137,8 +129,7 @@ const moveSuccessfulDrop = computed<boolean>(
   () =>
     imagePreloadingCompleted.value &&
     Boolean(mintingSession.value.items.length) &&
-    Boolean(mintingSession.value.txHash) &&
-    props.action.details.status === TransactionStatus.Finalized,
```
Jarsen136, line 141: Is it safe to delete this condition? What if the TX failed?

tests/randomize.spec.ts
```diff
@@ -1,12 +1,18 @@
-import { uidCrypto, uidMathDate } from '../utils/randomize'
+import {
+  uidCrypto,
+  // uidMathDate,
```
Jarsen136, line 3:
```suggestion
```

components/collection/drop/modal/PaidMint.vue
hassnian: > This is the line on which I need feedback. Sometimes, the status was never changed to `Finalized`

you mean the transaction was successful, but status was never updated to `Finalized` ?

preschian: > If possible, there are two TX with the same SN would be conflicted

Oh, this one is easier. Let me check

> you mean the transaction was successful, but status was never updated to Finalized ?

It seems like successful because we can safely close the tabs, and later on, the item will be updated

preschian: So far, it looks good if the transaction fails. It will blocked to the next step

Screenshot 2024-05-29 at 19 22 16

tests/randomize.spec.ts
preschian: updated

prury:
@preschian sidebar.spec.ts is failing on other PRs as well, investigating.

preschian:
> sidebar.spec.ts is failing on other PRs as well, investigating.

I tried to increase the timeout in the playwright global config. Should we increase it again? currently 12s. On local, `sidebar.spec.ts` looks fine

prury:
> > sidebar.spec.ts is failing on other PRs as well, investigating.
> 
> I tried to increase the timeout in the playwright global config. Should we increase it again? currently 12s. On local, `sidebar.spec.ts` looks fine

uh, not sure if i understood
from what I've seen you have increased the global test timeout from 1 min to 2 min (1 * 60 * 1000 to 2 * 60 * 1000), and retries from 1 to 3

i would leave the global test timeout at 1 min, meaning that, if a single test takes more than one minute, it would automatically fail(i try to keep each test below 1min)

retries i would also leave at 1, otherwise the test will take a long time to fail (imagine if a test takes 50s and it retries it 3 times)

if you were willing to change the expect timeout(5s default) we have to use: `config = { expect: { timeout: 10000 } }`

preschian:
> from what I've seen you have increased the global test timeout from 1 min to 2 min (1 * 60 * 1000 to 2 * 60 * 1000), and retries from 1 to 3

Oh sorry, yes this one correct. Not 12s but to 2 min

> if you were willing to change the expect timeout(5s default) we have to use: `config = { expect: { timeout: 10000 } }`

Yes, this is the one I add. Right now, the config:
```js
const TIMEOUT = 2 * 60 * 1000

const config = {
	// ...
	timeout: TIMEOUT,
	expect: {
	  timeout: TIMEOUT,
	},
	// ...
}
```

playwright.config.ts
```diff
@@ -22,7 +24,10 @@ export default defineConfig({
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
-  timeout: 1 * 60 * 1000,
+  timeout: TIMEOUT,
+  expect: {
+    timeout: TIMEOUT,
```
prury, line 29:
```suggestion
    timeout: 12000,
```

prury: if we change the expect timeout to 2min, every single expect assertion can take up to 2 min.
changed it to 12s as i think it was what you were trying to achieve
let's change it and run CI tests to see

prury:
as i said in dm, tests were failing due to an RPC error, that's why it failed across many PRs
this RPC error prevented the app from getting balances, making the add funds button never to reach load state on an empty account

feel free to revert config or timeout changes in tests, or test the 12s global timeout

Jarsen136 (approved):

hassnian (approved):
the only thing is that now we won't be able to see NFT previews while mass-minting , if @exezbcz approves let's go

exezbcz:
hmm weird, the modal does not even load for me 
image



image

exezbcz:
@hassnian that is the thing, we dont show any previews because what you get is random now

preschian:
> hmm weird, the modal does not even load for me

let me check

playwright.config.ts
preschian: I just reverted e2e stuff, since it's not related to this PR

preschian:
> hmm weird, the modal does not even load for me

interesting, how about now @exezbcz , probably due to error on rpc side

playwright.config.ts
prury: thank you

prury (approved):

Issue #10307, opened 2024-05-16
.hdd is very slow


JustLuuuu:
Drop
---
[.hdd](https://kodadot.xyz/ahp/drops/hdd) loading is extremely slow - at least the first variations. It shouldn't be like this

https://github.com/kodadot/nft-gallery/assets/104380776/3c3dae6f-00fb-4fe1-b8ac-d04867e7e195

Code-checker
---
The art itself in code-checker is much much faster
Code: 
[kodahash-main-hdd3.2.2.zip](https://github.com/kodadot/nft-gallery/files/15332216/kodahash-main-hdd3.2.2.zip)

https://github.com/kodadot/nft-gallery/assets/104380776/da022941-8609-4eb9-a802-bd224e947cdf

Why is it so slow if the code provided by artists is quick? this should not happen.

JustLuuuu:
looks like its all drops not only .hdd

https://github.com/kodadot/nft-gallery/assets/104380776/0c5768c9-bc16-4bbd-ab74-1bccf08d6aa3

springshine0925:
👋

PR #10379, opened 2024-06-03
perf: improve drop first load speed

hassnian:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- fetching the drop a little before inside the `setup` 
- made some components lazy 
- drop gallery item is loaded once the divider is seen  

- [x] Ref #10307

## Needs QA check

## Screenshot 📸

### DIFF
--- a/components/carousel/CarouselTypeRelated.vue
+++ b/components/carousel/CarouselTypeRelated.vue
@@ -1,10 +1,8 @@
 <template>
   <div>
-    <div ref="target"></div>
-    <CarouselIndex
-      v-if="nfts && targetIsVisible"
-      :title="`${$t('nft.related')}`"
-      :nfts="nfts" />
+    <LoadLazly>
+      <CarouselIndex v-if="nfts" :title="`${$t('nft.related')}`" :nfts="nfts" />
+    </LoadLazly>
   </div>
 </template>
 
@@ -16,7 +14,4 @@ const props = defineProps<{
 }>()
 
 const { nfts } = await useCarouselRelated({ collectionId: props.collectionId })
-
-const target = ref<HTMLHtmlElement>()
-const targetIsVisible = useOnceIsVisible(target)
 </script>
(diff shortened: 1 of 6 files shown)


Jarsen136 (approved):
You have applied the similar logic of lazy load in two places. How about making a common lazy wrapper component?


```
<div>
  <div ref="target"></div>
  <template v-if="targetIsVisible">
      <slot />
   </template>
<div>

const target = ref<HTMLHtmlElement>()
const targetIsVisible = useOnceIsVisible(target)
```


> drop gallery item is loaded once the divider is seen

In case of one day we removed the divider line on the drop page.

hassnian:
> You have applied the similar logic of lazy load in two places. How about making a common lazy wrapper component?

done

components/carousel/CarouselTypeRelated.vue
```diff
@@ -1,10 +1,8 @@
 <template>
   <div>
-    <div ref="target"></div>
-    <CarouselIndex
-      v-if="nfts && targetIsVisible"
-      :title="`${$t('nft.related')}`"
-      :nfts="nfts" />
+    <LoadLazly>
```
Jarsen136, line 3: Maybe `LoadLazily` or `LazyLoad`?

components/collection/drop/GenerativeLayout.vue
```diff
@@ -52,20 +52,22 @@
...
 
-      <hr class="my-20" />
-
-      <LazyCollectionDropItemsGrid
-        v-if="drop?.collection"
-        class="mb-14"
-        :collection-id="drop?.collection" />
+      <hr ref="divider" class="my-20" />
+
+      <LoadLazly :target="divider">
```
Jarsen136, line 57: Is it necessary to pass the `divider` here?

components/carousel/CarouselTypeRelated.vue
hassnian: done

components/collection/drop/GenerativeLayout.vue
hassnian: >Is it necessary ? 

no, but I do want the divider to be the target in this case , this will make the data load a little earlier

Jarsen136 (approved):

preschian (approved):
quite really difficult to see the differences. but, lets go

PR #10469, opened 2024-06-13
fix: Prevent user from adding comma on drops

### LINKED ISSUE #10465: Prevent user from adding comma on drops
image

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Closes #10465

## Screenshot 📸

### DIFF
--- a/components/collection/drop/MintStepper.vue
+++ b/components/collection/drop/MintStepper.vue
@@ -34,8 +34,4 @@ const max = computed(() => {
 
   return dropMax
 })
-
-watch(amountToMint, (amount) => {
-  amountToMint.value = (amount as number | '') === '' ? 1 : Math.ceil(amount)
-})
 </script>
--- a/components/collection/drop/MintStepper.vue
+++ b/components/collection/drop/MintStepper.vue
@@ -34,4 +34,8 @@ const max = computed(() => {
 
   return dropMax
 })
+
+watch(amountToMint, (amount) => {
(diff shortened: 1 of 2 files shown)


hassnian (commented):
I'd make this part of `NeoStepper`

Jarsen136:
> I'd make this part of `NeoStepper`

Updated

hassnian:
@Jarsen136 let's add a pattern to allow only numbers

CleanShot 2024-06-14 at 10 45 54@2x

Jarsen136:
> @Jarsen136 let's add a pattern to allow only numbers
> 
> CleanShot 2024-06-14 at 10 45 54@2x

As we have set `type="number"` on the input oruga component, It should be handled properly. With chrome browser, it does not allow me to type any special character. 

Then I tried to reproduce the bug and found it could happen on Safari. I can't find a simple way to detect if there is any special char. The `modal-value` doesn't return the special char in this case.

Jarsen136:
> Still, I managed to get "" so let's add 0 in that case, can we fix that, and I think we are good

I've thought about this before, but it's not appropriate. Imagine someone wants to delete all the entered numbers, set it to “”, and then enter “11”. The input box will always show ”0“  @hassnian 

Let's come back to the original issue. It fixes the problem of inputting non-integer with comma(,)/dot(.)

hassnian (approved):

preschian (approved):

PR #10555, opened 2024-06-30
feat: advance verify for profile changes

Jarsen136:

## PR Type

- [x] Feature

## Needs QA check

## Context

- [x] Related https://github.com/kodadot/private-workers/issues/184
- [x] Requires deployment https://github.com/kodadot/private-workers/pull/192

## Screenshot 📸


preschian (approved):

composables/useVerifyAccount.ts
```diff
@@ -30,36 +33,50 @@ export default function useVerifyAccount() {
...
-      walletStore.setSignedMessage(signature)
       return signature
     }
 
     throw new Error('You have not completed address verification')
   }
 
-  const getSignaturePair = async () => {
-    const signature = await getSignedMessage()
+  const getCustomSignaturePair = async (message) => {
```
hassnian, line 53:
```suggestion
  const getCustomSignaturePair = async (message: string): SignaturePair => {
```

composables/useVerifyAccount.ts
```diff
@@ -30,36 +33,50 @@ export default function useVerifyAccount() {
     return signedMessage
   }
 
-  const getSignedMessage = async () => {
+  const getSignedMessage = async (message) => {
```
hassnian, line 36:
```suggestion
  const getSignedMessage = async (message: string) => {
```
plus type the return

composables/useVerifyAccount.ts
```diff
@@ -12,7 +12,10 @@ const signMessagePolkadot = async (address: string, message: string) => {
   return signedMessage.signature
 }
 
-export const SIGNATURE_MESSAGE = 'Verify ownership of this account on Koda'
+const SIGNATURE_MESSAGE = 'Verify ownership of this account on Koda'
+
+export const generateSignatureMessageWithProfileVersion = (version: number) =>
```
hassnian, line 17: we might add the version system for other things , let's keep it generic here 

something like `generateVersionedSignatureMessage` or anything similar 

```suggestion
export const generateVersionedSignatureMessage = (version: number): string  =>
```

composables/useVerifyAccount.ts
```diff
@@ -30,36 +33,50 @@ export default function useVerifyAccount() {
...
+    if (signedMessage.value) {
+      return {
+        signature: signedMessage.value,
+        message: SIGNATURE_MESSAGE,
+      }
+    }
+    const signature = await getSignedMessage(SIGNATURE_MESSAGE)
+    if (signature) {
+      walletStore.setSignedMessage(signature)
+    }
```
hassnian, line 71: I don't think `signature` can be null here, but if that is the case

```suggestion
    const signature = await getSignedMessage(SIGNATURE_MESSAGE)
    if (!signature) {
      return null // ? or undefined 
    }
    
    walletStore.setSignedMessage(signature)
```

Jarsen136: If the user cancels the wallet signature process, it will throw error inside `getSignedMessage`. Therefore there's not need to throw an error again.

composables/useVerifyAccount.ts
Jarsen136: I have changed the name. However, I do not prefer to add the version system for other things, as the version is stored in the profile. It should be only used with profile-related.

vikiival:
cc @hassnian

hassnian (approved):

hassnian:
@Jarsen136 you have some conflicts

vikiival:
CC @Jarsen136

Jarsen136:
I didn't realize it was automatically merged...

It depends on changes on the worker side, and when it is released online, we should also pay attention to releasing the worker at the same time. 

https://github.com/kodadot/private-workers/pull/192 @vikiival @hassnian

PR #10626, opened 2024-07-15
:zap: Create is back

hassnian:
This reverts commit c716ffd1c71329c3d342075eb2a70887bc440091, reversing changes made to eb9f53e2649266dd7c34fcf777282a9588c8e3ca.

PR #10627, opened 2024-07-15
chore: use eslint modules

preschian:

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] Use eslint-modules. and unify eslint config. ref: https://eslint.nuxt.com/

will fix reviewdog in seperate PR


vikiival:
I've seen that you have used biome in one project, isn't it more suitable?

preschian:
> I've seen that you have used biome in one project, isn't it more suitable?

afaik biome still doesn't fully support some front-end frameworks

At the moment, for vue project -> eslint
for non-vue project -> biome. But I'm okay with switching to eslint so that we can use eslint in all our repos if you prefer eslint. Since biome config still not merged yet: https://github.com/kodadot/private-workers/blob/feat-dyndata/services/dyndata/biome.json

vikiival:
>  But I'm okay with switching to eslint so that we can use eslint in all our repos if you prefer eslint

I am ok with both, just seen on twitter that @biomejs is faster ;D

Jarsen136 (commented):
image

Is it possible to auto-fix these warnings related to eslint ?

preschian:
> Is it possible to auto-fix these warnings related to eslint ?

mostly able to auto-fix that: https://github.com/kodadot/nft-gallery/pull/10630
but some of them need manual fixes: https://github.com/kodadot/nft-gallery/issues/10631

components/common/successfulModal/ShareSocialsSection.vue
```diff
@@ -4,30 +4,61 @@
       :class="[
         `flex justify-around px-8 items-center w-full gap-6`,
         { '!gap-4': withCopy },
-      ]">
+      ]"
+    >
```
vikiival, line 8: this is very strange rule imo

vikiival (commented):
~

components/common/successfulModal/ShareSocialsSection.vue
preschian: it's customizable, but I prefer leave it as it is.

ref:
`⚙️ This rule is included in all of "plugin:vue/vue3-strongly-recommended"...`: https://eslint.vuejs.org/rules/html-closing-bracket-newline

vikiival (commented):
Will merge after 

- #10626

Otherwise waiting from review from @kodadot/internal-dev

hassnian (approved):

vikiival:
c @Jarsen136

Jarsen136 (approved):

vikiival (approved):

PR #10668, opened 2024-07-22
fix: Remove minted out collection from landing

### LINKED ISSUE #10664: Remove minted out collection from landing
Here, you can see on landing that Tilescapes are minted out. I would filter it in a way that if the collection is mint out, it will not be shown on the landing anymore - instead of that other collection from the list can take its place.

Screenshot 2024-07-22 at 12 20 09

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Closes #10664

## Screenshot 📸


components/drops/useDrops.ts
```diff
@@ -48,28 +49,33 @@ const DROP_LIST_ORDER = [
...
     const formattedDrops = await Promise.all(
-      dropsList.value.map(async (drop) => getFormattedDropItem(drop, drop)),
+      dropsList.value.map(async drop => getFormattedDropItem(drop, drop)),
     )
 
     drops.value = formattedDrops
 
+    if (filterOutMinted) {
+      drops.value = drops.value.filter(drop => !drop.isMintedOut)
+    }
```
hassnian, line 70: nitpick , this avoids updating the ref twice

``````
  let formattedDrops = await Promise.all(
      dropsList.value.map(async drop => getFormattedDropItem(drop, drop)),
    )

    if (filterOutMinted) {
      formattedDrops = formattedDrops.filter(drop => !drop.isMintedOut)
    }

    drops.value = formattedDrops
``````

or maybe use a `then` and this way you don't have to use `let` and override

hassnian (commented):
would be nice if we could filter this out on the worker side 

then we wouldn't have to keep track of minted out drops and update the limit here

Jarsen136:
> would be nice if we could filter this out on the worker side
> 
> then we wouldn't have to keep track of minted out drops and update the limit here
> 
> 

Yes, I have already thought about it. The reason stop me from implementing it is that the worker side does not have the correct stats of `minted`.

components/drops/useDrops.ts
Jarsen136: updated

hassnian (approved):

vikiival (approved):

vikiival:
You have conflicts @Jarsen136

Issue #10708, opened 2024-07-30
Where I can see BASE Drops? 


yangwao:
I've logged with EVM wallet and see only these networks 

![image](https://github.com/user-attachments/assets/c1e7440e-1bcf-4fd9-b820-68da28dc1473)

When I click on drops, link goes to https://koda.art/ahp/drops

yangwao:
Seems here it works https://deploy-preview-10670--koda-canary.netlify.app/base/drops
BUT
I think we had more that one drop on BASE in past? 

https://deploy-preview-10670--koda-canary.netlify.app/base/drops

PR #10670, opened 2024-07-23
feat: Support minting (free) drops on EVM

hassnian:
## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

test drop : `/base/drops/quadz `

fix/add:
- [x] evm drop minting 
- [x] evm balance / multi-balance
- [x] disable auto-telport on evm
- [x] disable `add balance` on evm
- [x] disable `listing` on evm drop
- [x] enable base drops on landing and `/drops`
- [x] add multi vm landing strip drop flow 
- [x] add filter wallet selection by vm on specific routes (/drops/alias , item gallery) 
- 
- [x] Closes #10648
- [x] Closes #10688
- [x] Closes #10691
- [x] Closes #10695

#### Did your issue had any of the "$" label on it?

- [x] Fill up your DOT address: [Payout](https://kodadot.xyz/dot/transfer?target=16faLfsywwNATaEfbH2ah75dn6ZmctQWpMS5G4KFhbmj5hnD)

## Screenshot 📸

- [x] My fix has changed **something** on UI; 

CleanShot 2024-07-24 at 16 42 40@2x

CleanShot 2024-07-26 at 15 27 41@2x

CleanShot 2024-07-30 at 11 38 58@2x

CleanShot 2024-08-01 at 17 07 04@2x

CleanShot 2024-08-01 at 17 09 45@2x

PR #10720, opened 2024-08-01
fix: display all available chains & check wallet VM when buying / minting nft

### LINKED ISSUE #10796: Show base in wallet selector if polkadot wallet connected
## What

let's put a "base" to wallet selector even when you are connected with your polkadot wallet
image
- it will reduce the confusion of why the user cannot see all the chains there 

## Task
- add base to network selector 
	- i think base is a good start since on other chains we don't have much content
- when the user clicks on the network with the wallet connected, display reconnect popup 
- identify edge cases when the user can cancel the connection and somehow dodge the modal
- add polkadot assethub among evm selectors

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Related #10708
- [x] Closed #10796

Display all available chains no matter what kind of wallet(SUB/EVM) you are connected to.

## Screenshot 📸

image


hassnian:
changing this allows you to  buy (for example) and ahp nft with a evm wallet connected 

let's add this back once multi-accounts is implemented after #10670, also @vikiival mentioned something about changing that switch to change EVMs and not chains

preschian (approved):

hassnian:
@jarsen can you include #10796 in this pr ? 

> - when the user clicks on the network with the wallet connected, display reconnect popup
> - identify edge cases when the user can cancel the connection and somehow dodge the modal


that should fix

> changing this allows you to buy (for example) an ahp nft with a evm wallet connected

exezbcz:
I mean, it does switch chains, but adds even more friction to minting a drop 
image



this is me minting a base drop with a polkadot wallet connected 
- there has to be a reconnect either when switching the chain or when clicking mint

Jarsen136:
> @jarsen can you include #10796 in this pr ? likely after #10772.
> 
> > * when the user clicks on the network with the wallet connected, display reconnect popup
> > * identify edge cases when the user can cancel the connection and somehow dodge the modal
> 

I've thought the same as you. We could show all available chains on the selector. Whenever the users do something on-chain operation(buy/mint), we will check if the current chain matches the wallet. If not, it will show the reconnect wallet popup modal.

It would be the easiest way to handle multiple chains. cc @exezbcz

exezbcz:
@Jarsen136 yeah this way you are only delaying the reconnect modal to the last possible moment, i am for this. Right now there was the reconnect only when clicking the drop card

vikiival:
cc @hassnian

hassnian (dismissed):

hassnian:
@exezbcz s-works ?

composables/useTransaction.ts
```diff
@@ -151,32 +149,35 @@ const useExecuteTransaction = (options: TransactionOptions) => {
...
     }
 
-    execByVm({
-      SUB: () => {
-        ;(howAboutToExecute as SubstrateHowAboutToExecute)(
-          accountId.value,
-          (params as ExecuteSubstrateTransactionParams).cb,
-          arg,
-          {
+    doAfterCheckCurrentChainVM(() => {
```
hassnian, line 152: I get this approach but I'm not to sold on adding this inside `useTransaction`, I'd rather show the modal before the action has started.

maybe because this is simplest solution atm? 

<details><summary>see</summary>

just a visual examaple of where this would run after action has started ,

CleanShot 2024-08-15 at 10 54 00@2x

</details>



 @preschian wdyt

hassnian (changes requested):
spent some time doing qa , cc @exezbcz 

code has changed quite a bit, re-requesting @preschian review 

1. now that we have all chains, what should we do here, allow only the current vm wallet or both, same for the drop page

CleanShot 2024-08-15 at 10 25 21@2x

2. we should remove that modal here and allow reconnect inside the drop (already working)

CleanShot 2024-08-15 at 10 30 05@2x


3. create collection and create nft missing the switch modal 
CleanShot 2024-08-15 at 10 43 22@2x


4. wallet connect modal selects the a defualt vm chain after the wallet is connected, pls remove that , otherwise you can do stuff like this

minting a collection on base 

https://github.com/user-attachments/assets/6bd6bab7-207d-4c1f-b78b-2da36bfa9c96

5. unrelated this pr , drop page can't change prefix 

https://github.com/user-attachments/assets/3fe9ad46-d13b-4e94-8e28-654be1afc89e

Jarsen136:
> spent some time doing qa
> 
> 
Thanks for taking the time to test it!

> 
> 1. now that we have all chains, what should we do here, allow only the current vm wallet or both, same for the drop page
> 

> 2. we should remove that modal here and allow reconnect inside the drop (already working)
> 

I will wait for @exezbcz response as this is the existing interaction process.



> 3. create collection and create nft missing the switch modal
> 4. wallet connect modal selects the a defualt vm chain after the wallet is connected, pls remove that , otherwise you can do stuff like this
> 
> minting a collection on base
> 
I have just fixed it.
Kapture 2024-08-15 at 10 44 40



> 5. unrelated this pr (new issue pls) , drop page can't change prefix , it's kinda weird allowing to change the chain and changing it back
> 
>  CleanShot.2024-08-15.at.10.49.31.mp4

will create another issue.

composables/useTransaction.ts
Jarsen136: > I'd rather show the modal before the action has started.
> 
> maybe because this is simplest solution atm?

Yes, it's the simplest solution, then we don't need to add this check everywhere.

although I have added checks for most cases such as mint/drop/buy, I leave it here to make sure it is impossible to do on-chain operations with an incorrect wallet.

preschian:
> spent some time doing qa
> 
> code has changed quite a bit, re-requesting @preschian review

thanks

vikiival:
cc @exezbcz

exezbcz:
> I've thought the same as you. We could show all available chains on the selector. Whenever the users do something on-chain operation(buy/mint), we will check if the current chain matches the wallet. If not, it will show the reconnect wallet popup modal.

yeah this sounds like it will be the easiest

for a second i though making it dynamic based on where the user currently is but that would bring even more complexity


## so to answer:


> 1. now that we have all chains, what should we do here, allow only the current vm wallet or both, same for the drop page

lets show both, with the "current VM tab" preselected 
- being on base and hitting connect (randomly on any page) it will show the evm tab selected
- being on polkadot eco, it will have ahp tab selected



> 3\. we should remove that modal here and allow reconnect inside the drop (already working)

yeah lets delay the reconnect till the last possible moment

--------
thanks for testing @hassnian and thank you for the implementation @Jarsen136 !

Jarsen136:
> 
> ## so to answer:
> > 1. now that we have all chains, what should we do here, allow only the current vm wallet or both, same for the drop page
> 
> lets show both, with the "current VM tab" preselected
> 
> * being on base and hitting connect (randomly on any page) it will show the evm tab selected
> * being on polkadot eco, it will have ahp tab selected
> 
> > 3. we should remove that modal here and allow reconnect inside the drop (already working)
> 
> yeah lets delay the reconnect till the last possible moment
> 

✅  Updated

image

components/common/ConnectWallet/WalletTabs.vue
```diff
@@ -31,6 +31,6 @@ const walletTabs = computed<{ label: string, value: ChainVM }[]>(() => {
       label: 'Evm',
       value: 'EVM',
     },
-  ].filter(tab => props.preselected ? tab.value === props.preselected : true)
+  ]
```
hassnian, line 34: `walletTabs` doesn't need to be a computed anymore

hassnian (approved):

components/common/ConnectWallet/WalletTabs.vue
Jarsen136: updated

exezbcz:
what is left here?

preschian (approved):

PR #10789, opened 2024-08-11
fix: Party kit popup does not remember the state

### LINKED ISSUE #10744: Party kit popup does not remember the state
image
i get this popup everytime i enter the drop page and its becoming quite annoying. 

# Task
- this popup should be displayed till the user choose an option
- option is either accept (represented with enter the party or with a close button) or there is opt out, turning the feature off 
- when one of this is clicked, don't display it on any other drops and remember the choice

hassnian:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

1. modal close did not accept `party mode`
2. chatted with @exezbcz  and preferences should be kept between sessions

- [x] Closes #10744

## Needs Design check

- @exezbcz please review

## Screenshot 📸

https://github.com/user-attachments/assets/78105e9e-785c-43d6-87b3-ad50944d399a

### DIFF
--- a/components/collection/drop/PartyModal.vue
+++ b/components/collection/drop/PartyModal.vue
@@ -1,7 +1,7 @@
 <template>
   <transition name="fade">
     <div
-      v-if="isPartyModalOpen"
+      v-if="getHasUserNotSetPartyMode"
       class="fixed top-20 md:top-24 left-4 z-[1000] w-[calc(100%-32px)] md:w-[31.25rem] h-fit shadow-primary bg-background-color !p-6 gap-6 flex border items-center"
     >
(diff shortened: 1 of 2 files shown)


hassnian:
@exezbcz you might also be getting the modal every time if you disconnect your wallet, at this moment we are clearing the user settings after the wallet is disconnected, I think we should keep that, or another option would be to keep track of settings per wallet.

exezbcz:
yup, with @hassnian we discussed that it will be good to keep the settings between sessions. 


we don't have that advanced settings to really get an advantage from wallet specific settings or clear on disconnect

vikiival (approved):

PR #10925, opened 2024-09-01
fix: initial data load in useSubscriptionGraphql

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

For the composable `useSubscriptionGraphql`, the initial data fetching should not trigger `onChange` callback

### DIFF
--- a/composables/useSubscriptionGraphql.ts
+++ b/composables/useSubscriptionGraphql.ts
@@ -44,8 +44,10 @@ export default function ({
       const newResult = response.data as any
 
       if (!isEqual(newResult, lastQueryResult)) {
-        $consola.log(`[Graphql Subscription] New changes: ${JSON.stringify(newResult)}`)
-        onChange({ data: newResult })
+        if (lastQueryResult !== null) {
+          $consola.log(`[Graphql Subscription] New changes: ${JSON.stringify(newResult)}`)
+          onChange({ data: newResult })
+        }
         lastQueryResult = newResult
       }
     }


composables/useSubscriptionGraphql.ts
```diff
@@ -44,8 +44,10 @@ export default function ({
       const newResult = response.data as any
 
       if (!isEqual(newResult, lastQueryResult)) {
-        $consola.log(`[Graphql Subscription] New changes: ${JSON.stringify(newResult)}`)
-        onChange({ data: newResult })
+        if (lastQueryResult !== null) {
+          $consola.log(`[Graphql Subscription] New changes: ${JSON.stringify(newResult)}`)
+          onChange({ data: newResult })
+        }
```
hassnian, line 50: this is not how the previous implementation worked , `onChange` **was triggered** when fetching also for the **first time** 

see`Buy.vue` how it's clear that it depends on the first fetch to also trigger `onChange`, with this change the buy modal never loads

if for some reason you want this behaviour i would make it optional as. param, when needed, otherwise you might be breaking other code 

atm the polling waits 6 seconds before making the first fetch, I'd just fix that on this pr

CleanShot 2024-09-02 at 07 01 17@2x

 from #10911,  pls suggest if you have any better fix :)

https://github.com/kodadot/nft-gallery/blob/bec90e2456d33518a8069656bbc0847e9f528a3c/composables/useSubscriptionGraphql.ts#L61

hassnian (changes requested):

composables/useSubscriptionGraphql.ts
Jarsen136: Well, originally it was supposed to be used to monitor upcoming changes. Since there is a need to trigger the first request, let's keep it as it is.
Thanks for your explanation.

PR #10967, opened 2024-09-11
feat: Mass minting drops on evm

### LINKED ISSUE #10815: Mass minting drops on evm
seems like we dont have this enabled, can we have a look at this? 

image

design is the same as with polkadot

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

1. should work but I've not executed any transaction since I don't have a test drop to use 
2. `GENSOL_ABI` is not deleted since it's used in other places like burning, listing etc, should that also use the fetched `abi`?
 
- [x] Closes #10815
- [x] Needs worker:  https://github.com/kodadot/private-workers/pull/219

## Screenshot 📸

CleanShot 2024-09-11 at 14 28 11@2x

CleanShot 2024-09-13 at 10 00 57@2x


preschian (approved):
wfm

> GENSOL_ABI is not deleted since it's used in other places like burning, listing etc, should that also use the fetched abi?

yes, I would prefer fetch `abi` endpoint instead. we can address this on separate PR

hassnian:
merging since

CleanShot 2024-09-16 at 09 48 00@2x

and worked for everyone  

CleanShot 2024-09-13 at 10 00 57@2x

PR #10975, opened 2024-09-13
✨ middleware profile page redirect

### LINKED ISSUE #10682: Profile: Profile link
Just a random suggestion. Can we have https://kodadot.xyz/ahp/profile and https://koda.art/ahp/profile links to always lead to your profile - Profile that is connected.

roiLeo:
- close #10682 
- tested using `/ahp/profile` while connected
- fix missing ts types

### DIFF
--- a/middleware/redirects.global.ts
+++ b/middleware/redirects.global.ts
@@ -28,10 +28,7 @@ export default defineNuxtRouteMiddleware((route) => {
         val === `/${urlPrefix.value}/profile`,
       replaceValue: () => {
         const { accountId } = useAuth()
-        if (accountId.value) {
-          return `/${urlPrefix.value}/u/${accountId.value}`
-        }
-        return `/${urlPrefix.value}`
+        return accountId.value ? `/${urlPrefix.value}/u/${accountId.value}` : `/${urlPrefix.value}`
       },
     },
     {
--- a/middleware/redirects.global.ts
+++ b/middleware/redirects.global.ts
@@ -31,7 +31,7 @@ export default defineNuxtRouteMiddleware((route) => {
         if (accountId.value) {
           return `/${urlPrefix.value}/u/${accountId.value}`
         }
-        return `/${urlPrefix.value}/explore/collectibles`
+        return `/${urlPrefix.value}`
       },
     },
     {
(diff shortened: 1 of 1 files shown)


middleware/redirects.global.ts
```diff
@@ -15,36 +16,47 @@ export default defineNuxtRouteMiddleware((route) => {
...
   const paths = [
     {
-      cond: val => val === '/drops',
+      cond: (val: string) => val === '/drops',
       replaceValue: '/ahp/drops',
     },
     {
-      cond: val =>
+      cond: (val: string) =>
+        val.startsWith(`/${urlPrefix.value}`) && val.endsWith('profile'),
```
Jarsen136, line 28:
```suggestion
        val === `/${urlPrefix.value}/profile`,
```
how about it?

middleware/redirects.global.ts
```diff
@@ -15,36 +16,47 @@ export default defineNuxtRouteMiddleware((route) => {
...
+      replaceValue: () => {
+        const { accountId } = useAuth()
+        if (accountId.value) {
+          return `/${urlPrefix.value}/u/${accountId.value}`
+        }
+        return `/${urlPrefix.value}/explore/collectibles`
```
Jarsen136, line 34: Why redirect to the explore page? Maybe to landing page`/${urlPrefix.value}`

roiLeo, line 34:
```suggestion
        return `/${urlPrefix.value}`
```

middleware/redirects.global.ts
roiLeo: what about trailing slash? I don't think it will work as before

middleware/redirects.global.ts
```diff
@@ -15,36 +16,47 @@ export default defineNuxtRouteMiddleware((route) => {
...
     {
-      cond: val =>
+      cond: (val: string) =>
+        val === `/${urlPrefix.value}/profile`,
+      replaceValue: () => {
+        const { accountId } = useAuth()
+        if (accountId.value) {
+          return `/${urlPrefix.value}/u/${accountId.value}`
+        }
+        return `/${urlPrefix.value}`
```
hassnian, line 34: nit: since it's only one if else 
```suggestion
        return accountId.value ? `/${urlPrefix.value}/u/${accountId.value}` : `/${urlPrefix.value}`
```

hassnian (approved):

hassnian:
wfm

a test is failing 
CleanShot 2024-09-16 at 13 22 50@2x

Jarsen136 (approved):

Issue #11019, opened 2024-09-20
Offers: Cannot accept offer


JustLuuuu:
I created an offer for this NFT: https://kodadot.xyz/ahp/gallery/158-2407236917
From this wallet: 1NFfEH3yspdEgLnhZ5QWgb7B2z5LNCAE3HUo5nDxF5MqTcj
I accepted it from my  other wallet: 14yzMSVPZVnaTx8Pk6vsr8Y5G1LPzSzvTrAwhasjidVtHbLq

After clicking on accept I got this error message. I have no idea what is the issue.
<img width="437" alt="Screenshot 2024-09-20 at 11 32 05" src="https://github.com/user-attachments/assets/35ba5748-4ddd-421b-8655-f5c4176872c6">

Video
---

https://github.com/user-attachments/assets/e82e6111-5f4f-486b-839a-a45126d8745f

Jarsen136:
> I created an offer for this NFT: https://kodadot.xyz/ahp/gallery/158-2407236917 From this wallet: 1NFfEH3yspdEgLnhZ5QWgb7B2z5LNCAE3HUo5nDxF5MqTcj I accepted it from my other wallet: 14yzMSVPZVnaTx8Pk6vsr8Y5G1LPzSzvTrAwhasjidVtHbLq
> 
> After clicking on accept I got this error message. I have no idea what is the issue. 


This is due to insufficient balance. The offer price is 1 dot but the transferable balance in this account is 0.985 dot.

![image](https://github.com/user-attachments/assets/4237be2e-6660-4a14-a6b6-50482873b23c)

![image](https://github.com/user-attachments/assets/fc31d26b-9831-4804-823d-cccbe738b02e)

JustLuuuu:
@Jarsen136 how can I even create an offer on amount I dont have?

vikiival:
prob fixed in https://github.com/kodadot/nft-gallery/pull/11035 ?

Jarsen136:
> prob fixed in #11035 ?

In most cases, yes. 
But it is still possible that if a user transfers out all the balance in their account, others will not be able to accept his offer.

vikiival:
cc @JustLuuuu this should be done

PR #11035, opened 2024-09-23
fix: offer total amount

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Related #11019

The user balance should be higher than `offer price` plus `offer making cost 0.05DOT` 

## Screenshot 📸

image
image

### DIFF
--- a/composables/transaction/transactionOffer.ts
+++ b/composables/transaction/transactionOffer.ts
@@ -13,7 +13,7 @@ export const getOfferCollectionId = (prefix: Prefix) => {
   }
 }
 
-export const OFFER_MINT_PRICE = '500000000'
+export const OFFER_MINT_PRICE = 5e8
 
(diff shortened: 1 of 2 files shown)


composables/transaction/transactionOffer.ts
```diff
@@ -13,7 +13,7 @@ export const getOfferCollectionId = (prefix: Prefix) => {
   }
 }
 
-const OFFER_MINT_PRICE = '500000000'
+export const OFFER_MINT_PRICE = '500000000'
```
hassnian, line 16: can we do `5e8`, so this becomes a number by default and we can make it a string where needed

Jarsen136: updated

vikiival (approved):

hassnian (approved):

PR #11037, opened 2024-09-23
fix: Drop calendar: Collection with price that is not set should be TBA

### LINKED ISSUE #11034: Drop calendar: Collection with price that is not set should be TBA
When I don't set the price in the Cloudfare - the price in the calendar shows TBA. Which is good. But this TBA should be on both - the card and when you open the card (check pic). If its not possible to add it to card - let me know and I will try to set the price in all cases. Just sometimes we dont agree on it earlier

You can see that on collection **Alpinista**.

Screenshot 2024-09-23 at 12 44 09

Jarsen136:

## PR Type

- [x] Bugfix

## Needs QA check

## Context

- [x] Closes #11034

## Screenshot 📸

image

### DIFF
--- a/components/drops/BasicDropCard.vue
+++ b/components/drops/BasicDropCard.vue
@@ -59,7 +59,7 @@
                   </span>
                 </div>
               </slot>
-              <div v-if="dropPrice === null || dropPrice === ''">
+              <div v-if="isTBA(dropPrice)">
                 <span>TBA</span>
               </div>
               <div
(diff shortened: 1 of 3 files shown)


components/drops/calendar/DropPreviewModal.vue
```diff
@@ -210,5 +210,5 @@ const formattedDate = computed(() =>
 )
 
 const withPlaceholder = (nullableValue: unknown, or: unknown) =>
-  nullableValue === null ? placeholder : or
+  nullableValue === null || nullableValue === '' ? placeholder : or
```
hassnian, line 213:
```suggestion
  isTBA(nullableValue) ? placeholder : or
```

components/drops/BasicDropCard.vue
```diff
@@ -59,8 +59,11 @@
                   </span>
                 </div>
               </slot>
+              <div v-if="dropPrice === null || dropPrice === ''">
```
hassnian, line 62:
```suggestion
              <div v-if="isTBA(dropPrice)">
```

hassnian (approved):

vikiival (approved):

PR #11068, opened 2024-10-01
Trim trailing zeroes

### LINKED ISSUE #10230: duplicated zeros in decimals
image
It would be nice to make it consistent, the input shows 0,5 and the price is 0.5 

I am not sure if its automatic somehow, but i would prefer seeing 0.5 everywhere

uDaiko:
After I stumpled upon this, as discussed in https://github.com/kodadot/nft-gallery/issues/10230

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Needs QA check

## Context

- [x] Closes #10230
- [ ] Requires deployment <worker/stick/speck>

## Screenshot 📸

firefox_XHgWUfXpNE
firefox_kQGteIoil3


components/offer/MakingOfferSingleItem.vue
```diff
@@ -88,10 +89,23 @@ const offerExpirationStoreItem = computed({
...
 )
+const trimmedCollectionFloorPrice = computed(() => trimDuplicateZeroes(collectionFloorPrice.value))
+
+const trimDuplicateZeroes = (price: string): string => {
+  if (price === '--') {
+    return '--'
+  }
+  const [amount, chainSymbol] = price.split(' ')
+
+  const trimmedAmount = parseFloat(amount).toString()
```
Jarsen136, line 105: It's unnecessary to add another function and process the string. There is already `formatWithBlank` function. Could you handle it there? 

maybe
```
const formatWithBlank = (value: number) => {
  return value ?  `${parseFloat(formatBalance(value, decimals.value))} ${chainSymbol.value}`: '--'
}

```

Jarsen136 (changes requested):

components/offer/MakingOfferSingleItem.vue
uDaiko: You are right, it's just a wrapper function that would be way cleaner. I'll change that

Jarsen136:
I just rechecked #10230 and found that it happened when you listed some NFT. Could you also fix it there?

image

uDaiko:
@Jarsen136 good catch. Aside from `MakingOfferSingleItem` and with your Input I found `ListingCartSingleItemCart` and `ListingCartItem` where that issue appears. Given that all these basically have the same requirement I thought it would be better to just refactor it and put the parse Code into the balance util. Please have a look.

Jarsen136 (approved):

components/offer/MakingOfferSingleItem.vue
```diff
@@ -87,10 +84,10 @@ const offerExpirationStoreItem = computed({
   set: v => offerStore.updateItem({ id: item.value.id, offerExpiration: v }),
 })
 
-const highestOfferPrice = computed(() => formatWithBlank(Number(item.value.highestOffer) || 0) || '--')
+const highestOfferPrice = computed(() => formatBalanceWithBlank(Number(item.value.highestOffer) || 0))
```
hassnian, line 87: missing decimals  and chain symbol

utils/format/balance.ts
```diff
@@ -88,6 +88,10 @@ export function roundTo(value: number | string, limit = 2) {
   })
 }
 
+export function formatBalanceWithBlank(value: number, decimals?: number, chainSymbol?: string) {
+  return value ? `${parseFloat(format(value, decimals))} ${chainSymbol}` : '--'
+}
```
hassnian, line 93: we already have utils for formatting amounts pls use those 

have a look at :
- functions `formatAmountWithRound` or `roundTo` inside `utils/format/balance.ts`
- composable `useAmount`

I'd add a blank param to those

utils/format/balance.ts
```diff
@@ -88,6 +88,10 @@ export function roundTo(value: number | string, limit = 2) {
   })
 }
 
+export function formatBalanceWithBlank(value: number, decimals?: number, chainSymbol?: string) {
```
hassnian, line 91: fyi: even though format has default `decimals` of `12` but not all chains have the same value

hassnian (changes requested):

vikiival:
cc @uDaiko

uDaiko:
@hassnian I used the function you mentioned, please have a look. It's working however I noticed that leaving the  roundBy parameter blank as you suggested results in unreliable rounds from 0.3999 to 0.4. Would probably make sense to just add  4 round to the function calls

components/common/listingCart/multipleItemsCart/ListingCartItem.vue
```diff
@@ -57,7 +56,7 @@ const props = defineProps<{
 }>()
 
 const floor = computed(() =>
-  formatBalance(props.nft.collection.floor, decimals.value, chainSymbol.value),
+  useAmount(computed(() => props.nft.collection.floor), decimals, chainSymbol).formatted,
 )
```
hassnian, line 60:
```suggestion
  const { formatted: floor } = useAmount(computed(() => props.nft.collection.floor), decimals, chainSymbol)
```

components/offer/MakingOfferSingleItem.vue
```diff
@@ -73,10 +72,12 @@ const { chainSymbol, decimals } = useChain()
...
 
-const itemPrice = computed(() => formatWithBlank(Number(item.value.price)))
+const itemPrice = computed(() => formatWithBlank(item.value.price))
 
-const formatWithBlank = (value: number) => {
-  return value ? formatBalance(value, decimals.value, chainSymbol.value) : '--'
+const formatWithBlank = (value: string | undefined) => {
+  return value
+    ? useAmount(computed(() => value), decimals, chainSymbol).formatted
+    : '--'
```
hassnian, line 80: DRY lets avoid this code duplication

I was suggesting something like 

````

export default function (
  tokenAmount: ComputedRef<number | string | undefined>,
  tokenDecimals: ComputedRef<number>,
  chainSymbol: ComputedRef<string>,
  { roundBy, withBlank = false }: { roundBy?: ComputedRef<Prefix | number>, withBlank?: boolean } = {}
) {
  const { getCurrentTokenValue } = useFiatStore()

  const amountFormatted = computed(() => {
    const amount = tokenAmount.value
      ? formatAmountWithRound(tokenAmount.value, tokenDecimals.value, roundBy?.value)
      : 0

    if (amount && withBlank) {
      return '--'
    }

    return `${amount} ${chainSymbol.value}`
  })

  const amountUsd = computed(() => {
    const value = calculateUsdFromToken(
      Number(tokenAmount.value) * Math.pow(10, -tokenDecimals.value),
      Number(getCurrentTokenValue(chainSymbol.value)),
    )
    return typeof value === 'number' && value >= 0 ? `$${value}` : ''
  })

  return {
    formatted: amountFormatted,
    usd: amountUsd,
  }
}

````

so you can do 

```suggestion
  return useAmount(computed(() => value), decimals, chainSymbol, { withBlank: true } ).formatted.value
```

components/offer/MakingOfferSingleItem.vue
```diff
@@ -87,11 +88,9 @@ const offerExpirationStoreItem = computed({
   set: v => offerStore.updateItem({ id: item.value.id, offerExpiration: v }),
 })
 
-const highestOfferPrice = computed(() => formatWithBlank(Number(item.value.highestOffer) || 0) || '--')
+const highestOfferPrice = computed(() => formatWithBlank(item.value.highestOffer))
```
hassnian, line 91: and you don't even need `formatWithBlank` just do

```suggestion
const {formatted: highestOfferPrice}  = useAmount(computed(() => value), decimals, chainSymbol, { withBlank: true } )
```

components/common/listingCart/singleItemCart/ListingCartSingleItemCart.vue
```diff
@@ -72,7 +71,9 @@ const collectionPrice = computed(() =>
 )
 
 const formatWithBlank = (value: number) => {
-  return value ? formatBalance(value, decimals.value, chainSymbol.value) : '--'
+  return value
+    ? useAmount(computed(() => value), decimals, chainSymbol).formatted
+    : '--'
```
hassnian, line 76: use composable as mentioned above

hassnian (changes requested):

hassnian:
> I noticed that leaving the roundBy parameter blank as you suggested results in unreliable rounds from 0.3999 to 0.4.

that's fine, if for some reason you don't need the rounding and you need to show the exact number 

you can add a optional logic to `roundTo` by adding `minimumFractionDigits`

components/offer/MakingOfferSingleItem.vue
uDaiko: I'm sure you meant `if (!amount && withBlank) {
      return '--'
    }` ?
    Doesn't it make more sense to just check for amount falsy and additional blank option and just return '--' straight up and then proceed

components/offer/MakingOfferSingleItem.vue
uDaiko: I personally think having a wrapper function to handle the function call with passing parameters, options and then just pass the value of whatever price property is doing it right now to it is cleaner than doing the whole full function call 3 or 2 times. But sure I can do it like that if that's the preferred way.

components/offer/MakingOfferSingleItem.vue
hassnian: > I'm sure you meant `if (!amount && withBlank) { return '--' }` ?

yes mb

> Doesn't it make more sense to just check for amount falsy and additional blank option and just return '--' straight up and then proceed

you don't want to have '--' in all instaces, sometimes you do want to show that value is '0 DOT' for example

uDaiko:
> > I noticed that leaving the roundBy parameter blank as you suggested results in unreliable rounds from 0.3999 to 0.4.
> 
> @uDaiko that's fine, if for some reason you don't need the rounding and you need to show the exact number
> 
> you can add a optional logic to `roundTo` by adding `minimumFractionDigits`
> 

I was thinking of just passing `useAmount(computed(() => props.nft.collection.floor), decimals, chainSymbol, computed(() => 4)`
alternatively you can have roundTo to have a default of 4 unless stated otherwise  but then existing calls of this function would have to be checked and modified so I think it's easier to just pass it in the 3 instances

components/offer/MakingOfferSingleItem.vue
hassnian: if you go with util function move it inside a separate file like `/utils/format/balance.ts` but dont use a composable inside 

or use composables directly

components/offer/MakingOfferSingleItem.vue
uDaiko: oh yeah of course that's what the blank option is for. Might have worded that wrong.

composables/useAmount.ts
```diff
@@ -5,15 +5,15 @@ export default function (
...
+  { roundBy, withBlank = false }: { roundBy?: ComputedRef<Prefix | number>, withBlank?: boolean } = {},
 ) {
   const { getCurrentTokenValue } = useFiatStore()
 
   const amountFormatted = computed(() => {
     const amount = tokenAmount.value
       ? formatAmountWithRound(tokenAmount.value, tokenDecimals.value, roundBy?.value)
       : 0
-    return `${amount} ${chainSymbol.value}`
+    return ((!amount || amount === '0') && withBlank) ? '--' : `${amount} ${chainSymbol.value}`
```
hassnian, line 16:
```suggestion
    return !Boolean(Number(amount)) && withBlank ? '--' : `${amount} ${chainSymbol.value}`
```

make the conditional checking a util inside the composable and also uses it inside `amountUsd`



````

const amountWithBlank = (amount) => 

````

hassnian (commented):
almost there!

hassnian (approved):
there are a couple of edge cases that need covering, let's keep it simple 

reverted to a simpler version 405b0f7     

@Jarsen136 code changed a bit pls re-review ty

Jarsen136 (approved):

vikiival:
Thank u @uDaiko !

PR #11086, opened 2024-10-07
fix: Evm wallet chain not reacting to prefix change

### LINKED ISSUE #11024: Evm wallet chain not reacting to prefix change
### What happened?

CleanShot 2024-09-21 at 14 53 10@2x

hassnian:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] Closes #11024

## Screenshot 📸

https://github.com/user-attachments/assets/53b82e8c-da75-4e2b-8248-166ccc93d4a4

### DIFF
--- a/composables/transaction/evm/useMetaTransaction.ts
+++ b/composables/transaction/evm/useMetaTransaction.ts
@@ -1,6 +1,6 @@
 import { type Address, TransactionExecutionError } from 'viem'
 import { simulateContract, waitForTransactionReceipt, writeContract } from '@wagmi/core'
-import { useConfig } from '@wagmi/vue'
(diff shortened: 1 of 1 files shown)


hassnian:
@exezbcz what about that icon, now that is not synced with the prefix 

so you can be on`mnt` prefix and see `base` as the wallet's chain

I think It makes sense, wdyt

CleanShot 2024-10-07 at 08 45 39@2x

vikiival:
> I think It makes sense, wdyt

Would use VM based prefixes?

vikiival (approved):

PR #11113, opened 2024-10-17
feat: Control return to confirm action

### LINKED ISSUE #10213: Control return to confirm action

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

continue #10306

@Jarsen136 let me know if you find anything, this approach seems to fix the issues that had previously and it's simpler imo

- [x] Closes #10213

## Needs Design check

- @exezbcz please review

## Screenshot 📸

CleanShot 2024-10-17 at 12 22 44

### DIFF
--- a/composables/useKeyboardKeys.ts
+++ b/composables/useKeyboardKeys.ts
@@ -1,0 +1,11 @@
+import { useMagicKeys } from '@vueuse/core'
+
+export function useKeyboardKeys({ onPressControlEnter }) {
+  const { Meta, Control, Enter } = useMagicKeys()
+
+  watch([Meta, Control, Enter], () => {
+    if (onPressControlEnter && (Meta.value || Control.value) && Enter.value) {
+      onPressControlEnter()
+    }
+  })
+}
--- a/composables/useKeyboardKeys.ts
+++ b/composables/useKeyboardKeys.ts
@@ -1,11 +1,15 @@
 import { useMagicKeys } from '@vueuse/core'
 
-export function useKeyboardKeys({ onPressControlEnter }) {
+export function onPressControlEnter(callback: () => void) {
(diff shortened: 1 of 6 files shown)


composables/useKeyboardKeys.ts
```diff
@@ -0,0 +1,15 @@
...
+  return watch([Meta, Control, Enter], () => {
+    if ((Meta.value || Control.value) && Enter.value) {
+      callback()
+    }
+  })
+}
+
+export function useKeyboardKeys({ onPressControlEnter: handlePressControlEnter }) {
+  handlePressControlEnter && onPressControlEnter(handlePressControlEnter)
+}
```
Jarsen136, line 15:
```suggestion
function handlePressControlEnter(callback: () => void) {
  const { Meta, Control, Enter } = useMagicKeys()

  return watch([Meta, Control, Enter], () => {
    if ((Meta.value || Control.value) && Enter.value) {
      callback()
    }
  })
}

export function useKeyboardKeys({ onPressControlEnter }) {
  onPressControlEnter && handlePressControlEnter(onPressControlEnter)
}
```

Jarsen136 (commented):
otherwise looks ok to me

composables/useKeyboardKeys.ts
hassnian: I'd rather have `onPressControlEnter` as a helper method like `onModalAnimation`

`useKeyboardKeys` is not what I'm using

hassnian: done f8b92a7


CleanShot 2024-10-19 at 07 07 12

Jarsen136 (approved):

PR #11119, opened 2024-10-19
feat: code checker more validation

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

Arkain faced some issues doing his drop 

- add: check if `p5.min.js` file is present, will show 
CleanShot 2024-10-19 at 10 25 03@2x
- add: check if `postMessageKoda` is called , will show 

CleanShot 2024-10-19 at 10 30 25@2x

@JustLuuuu would you want a separate check ? or is this clear enough? or we can change the copy to include 

postMessageKoda() is not called at the end of the draw() function

````
function draw() {
  background(bg)
  // Once the drawing is done, we can send the image to the parent window
  postMessageKoda()
}
````

### DIFF
--- a/components/codeChecker/CodeChecker.vue
+++ b/components/codeChecker/CodeChecker.vue
@@ -279,8 +279,7 @@ const onFileSelected = async (file: File) => {
   clear()
   startClock()
   selectedFile.value = file
-  const { indexFile, sketchFile, entries }
-    = await extractAssetsFromZip(file)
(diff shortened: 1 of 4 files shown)


vikiival:
cc @JustLuuuu

preschian (approved):

JustLuuuu:
I think this should be enough!

JustLuuuu:
wfm

vikiival (approved):

PR #11157, opened 2024-10-22
perf: Improve `collectionForMint` query

### LINKED ISSUE #7298: Improve `collectionForMint` query
in the existing code:
- no limit
- map and filter, what happen if the nfts.count more than > 1k

https://github.com/kodadot/nft-gallery/blob/9b79ef3e1477388686781db8fe8450a255a56643/components/create/CreateNft.vue#L303-L322

hassnian:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

## Context

- [x] perf: remove over fetching of nfts of each collection
- [x] Closes #7298

## Screenshot 📸


composables/massmint/useMassMint.ts
```diff
@@ -34,48 +34,67 @@ export const statusClass = (status?: Status) => {
...
+                collection: { id_eq: "${collection.id}" }
+                burned_eq: false
+            }) {
+                totalCount
+            }
+          `)
+            .then(response => resolve(response.data.nftEntitiesConnection.totalCount))
+            .catch(reject),
+        )))
         : null,
```
vikiival, line 79: I would rather write it as a block 

```js
if (!collectionEntities.value?.length) {
  return null
}

const ...
```

Moreover you know how many nfts was burned :) 

cause `supply` will change if you burn something 
but `nftCount` not

hassnian: thanks! wasn't aware of that field , updated 87ebd22

composables/massmint/useMassMint.ts
```diff
@@ -34,48 +34,41 @@ export const statusClass = (status?: Status) => {
...
     if (collectionEntities?.length) {
-      const collections = collectionEntities
-        .map(collection => ({
-          ...collection,
-          lastIndexUsed: Math.max(
-            ...collection.nfts.map(nft => Number(nft.index)),
-          ),
-
-          alreadyMinted: collection.nfts?.length,
-          totalCount: collection.nfts?.filter(nft => !nft.burned).length,
```
Jarsen136, line 66: Will they be used somewhere?

composables/useSubscriptionGraphql.ts
```diff
@@ -32,16 +32,7 @@ export default function ({
...
-        method: 'POST',
-        body: {
-          query: `
-          query {
-            ${query}
-          }
-        `,
-        },
-      })
+      const response = await fetchGraphql(query, { client })
```
Jarsen136, line 35: As the only use case, Is it necessary to move to the utils function?

composables/massmint/useMassMint.ts
hassnian: renamed the fields inside `collectionForMint.graphql`, so no need to map it again

CleanShot 2024-10-28 at 07 48 03@2x

composables/useSubscriptionGraphql.ts
hassnian: a prev version of the code was using it , removed it since it's not used anymore

vikiival (approved):

Jarsen136 (approved):

PR #11176, opened 2024-11-12
🔧 action list price is_null

### LINKED ISSUE #11175: Explore Items Listing Not working
### What happened?

either you are able to list and it's not working or you shouldn't be able to from this page and we should hide the "select" button

https://github.com/user-attachments/assets/2a9d7111-8e0c-458d-9674-bdf4edfe08ea

roiLeo:
- closes #11175 
- `price` is null, it's causing the bug since we do `price_eq`/`price_gt`: `0`

note: pain to debug


preschian (approved):

hassnian:
@roiLeo it's working for some NFTs and for some it's still not 

https://github.com/user-attachments/assets/502a898e-d443-4502-a22a-2565892072e6

roiLeo:
> @roiLeo it's working for some NFTs and for some it's still not

can't reproduce, could you share your address?

components/items/ItemsGrid/ItemsGridImageTokenEntity.vue
```diff
@@ -200,9 +200,8 @@ const onClickShoppingCart = async () => {
 const onClickListingCart = async () => {
   const nftsToProcess = await getTokensNfts([props.entity])
 
-  for (const nft of nftsToProcess) {
```
roiLeo, line 203: This part make a weird loop of the same items on some generative nfts

vikiival: cc @Jarsen136 ?

roiLeo: thanks for the clarification, i've added missing variables

components/items/ItemsGrid/useNftActions.ts
```diff
@@ -40,6 +40,12 @@ export const prepareListingQuery = (
         token: tokenSearchTerm,
         currentOwner_eq: accountId.value,
         burned_eq: false,
+        OR: {
+          price_isNull: true,
+          token: tokenSearchTerm,
+          currentOwner_eq: accountId.value,
+          burned_eq: false,
+        },
```
Jarsen136, line 48: I'm confused about this OR condition here. Would it be possible to remove the price limit (lines 53- 54) instead?

image

roiLeo: if there are no edge case Ok

Jarsen136 (approved):

PR #11204, opened 2024-12-02
feat: Reuse unused tokens for the next offer

### LINKED ISSUE #10963: Automatic cancel expired offer and return deposit
/// This function is used to cancel the specified swap offer created by the `caller` account. If
	/// the swap offer's deadline has not yet passed, the `caller` must be the owner of the offered
	/// item; otherwise, anyone can cancel an expired offer.
	/// After canceling the swap offer, the function emits the `SwapCancelled` event.
	///

- https://github.com/paritytech/polkadot-sdk/blob/9079f36b99ee0910762ce6020ca9dfd6671f77b3/substrate/frame/nfts/src/features/atomic_swap.rs#L106C1-L110C5

In human lang: 
If the offer is expired, we can setup a bot that will "clear" offers every N-hours

_Originally posted by @vikiival in https://github.com/kodadot/nft-gallery/issues/10951#issuecomment-2340393140_

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

> That has to be burned by user, or **you can reuse it for the next offer**

let's save this cost to the user, by reusing unsued tokens from old offers

- [x] Closes #10963

## Screenshot 📸


Jarsen136 (commented):
It's definitely a good idea to reuse the unused offer token. However, it need a better way to get the unused nft id. 

---
Let me provide an example showing how it may cause chaotic situations.

1. In the beginning you have an expired offer
 ```
[ { offerId: 1, nftSn: 1, status: EXPIRED }]
```


2. Then you try to create a new offer and reuse the unused token
``` 
[ { offerId: 1, nftSn: 1, status: EXPIRED },
{ offerId: 2, nftSn: 1, status: ACTIVE }]
```

3. If you continue to create a new offer
```
[ { offerId: 1, nftSn: 1, status: EXPIRED },
{ offerId: 2, nftSn: 1, status: ACTIVE },
{ offerId: 3, nftSn: 1, status: ACTIVE }]
```

In this step, there is already a problem occurred. You may create infinite offer with one nft.

4. More chaos is coming If the offer 2 has been accepted.
```
[ { offerId: 1, nftSn: 1, status: EXPIRED },
{ offerId: 2, nftSn: 1, status: ACCEPT },
{ offerId: 3, nftSn: 1, status: ACTIVE }]
```
If you continue to creating a new offer, the current implement will still using `nftSn: 1`.
However, the `nftSn: 1` is not available anymore.
This step would cause problems again.
```
[ { offerId: 1, nftSn: 1, status: EXPIRED },
{ offerId: 2, nftSn: 1, status: ACCEPT },
{ offerId: 3, nftSn: 1, status: ACTIVE }
{ offerId: 4, nftSn: 1, status: ACTIVE }]
```
---
Let me know if this example is not clear or it's not possible to happen with the current implement.

**Comment by hassnian** [comment]
> Let me know if this example is not clear or it's not possible to happen with the current implement.

the example you provided would defiantly be an issue if each new atomic swap mapped to an offer in our indexer

an offer's id is made of `{offeredCollectionId}-${offeredItem}` 

so when we create a new offer, with a reused token it will override the old offer [code](https://github.com/kodadot/stick/blob/2dd783ea31bfbf018381c80b9f71c30d62031136/src/mappings/nfts/createSwap.ts#L32)

`````
const id = createTokenId(event.collectionId, event.sn)
const final = offer ? await getOrCreate(context.store, Offer, id, {}) : await getOrCreate(context.store, Swap, id, {})
`````

to summarize as of now **you can't have multiple swaps/offers with the same offered item** 

either by design or a bug  cc @vikiival 

will make changes accordingly

**Comment by vikiival** [comment]
> so when we create a new offer, with a reused token it will override the old offer

Yup, that's how it is implemented 

https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/nfts/src/features/atomic_swap.rs#L82-L91

Swap is stored inside a map by key - `(offeredCollection, offeredItem)` - aka same as in our indexer (note: except that that key in rust is tuple and in our case is dash-separated string).

> so basically atm you can't have multiple swaps/offers with the same offered item

So yes each, one offered item per desired item

**Review by Jarsen136 — APPROVED** [review]

File: composables/transaction/transactionOffer.ts [hunk 1]
```diff
@@ -17,35 +19,44 @@ export const OFFER_MINT_PRICE = 5e8
...
-        collectionId,
-        nftSn,
+        offeredCollectionId,
+        offeredItem,
+        desiredCollectionId,
+        desiredItem,
         {
           amount: Number(price) || 0,
           direction: 'Send',
         },
```
vikiival, line 53: Make this null instesd of object if price is not provided

vikiival (approved):

composables/transaction/transactionOffer.ts
hassnian: atm users can't crate an offer with 0 price

made these changes to make sure is not possible to call the transaction without a price  [c8157cf](https://github.com/kodadot/nft-gallery/pull/11204/commits/c8157cfd70be373e91ec3afcaf476509eb3c1c2a)

or you mean they should be able to , "free" offer ? 



CleanShot 2024-12-29 at 12 05 27@2x

PR #11223, opened 2024-12-09
✨ profile Swaps button

### LINKED ISSUE #11214: Swaps in Profile Page follow up
add button to start a swap with that user

closed by #11223

CleanShot 2024-12-05 at 08 33 53@2x

### Swap Table

- show surcharge direction
- show both tokens that are being swapped , with the current implementation only the receiver or creator of the swap can open the modal, there is no `view swap` button, so users that are not involved in the trade can't see both tokens of the swap from this page
- also it should support swap entire collection 

image

### LINKED ISSUE #11216: Swap Landing link
how do you land on the Swap Page  `/prefix/swap`

maybe we can add it in the footer

CleanShot 2024-12-05 at 09 36 29@2x

roiLeo:
- closes #11214 
- closes #11216

### DIFF
--- a/components/profile/ProfileDetail.vue
+++ b/components/profile/ProfileDetail.vue
@@ -92,6 +92,14 @@
             @follow:fail="openProfileCreateModal"
             @unfollow:success="handleFollowRefresh"
           />
+          <NeoButton
+            :tag="NuxtLink"
+            variant="outlined-rounded"
+            icon-left="arrow-right-arrow-left"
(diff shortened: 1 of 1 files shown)


vikiival (approved):

hassnian (changes requested):
if you are not connected with a wallet from the same vm, you should show the vm switch modal


already implemented in `/drops/id`

CleanShot 2024-12-10 at 08 37 03@2x

CleanShot 2024-12-10 at 09 00 34@2x

hassnian (approved):

PR #11267, opened 2024-12-18
feat: Enhanced Bulk Minting: Add Support for Onchain Attributes

### LINKED ISSUE #11149: Enhanced Bulk Minting: Add Support for Onchain Attributes
Extend the current bulk minting functionality to allow users to add onchain attributes to NFTs during the minting process. These attributes should be stored on-chain and be available for querying and displaying.

Jarsen136:

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Supported description file type: json and txt. 
- CSV continued to https://github.com/kodadot/nft-gallery/issues/11282
- [x] Refactor

## Needs Design check

- @exezbcz please review

## Needs QA check

## Context

- [x] Closes #11149

## Screenshot 📸

image
image
image
image

minted nft example: https://canary.kodadot.xyz/ahk/gallery/462-9 \ https://canary.kodadot.xyz/ahk/gallery/462-10


hassnian (changes requested):
1. modal seems to be broken

CleanShot 2024-12-22 at 07 31 57@2x

2. attributes component not working if the item is not populated

https://github.com/user-attachments/assets/c9470841-cc04-4934-86c9-ff81a05c1c97

Jarsen136:
> 1. modal seems to be broken
> 
> 
> 2. attributes component not working if the item is not populated
> 

✅ All fixed

components/massmint/EditPanel.vue
```diff
@@ -100,6 +107,7 @@ import {
   NeoSidebar,
 } from '@kodadot1/brick'
 import type { NFT } from './types'
+import CustomAttributeInput from '@/components/rmrk/Create/CustomAttributeInput.vue'
```
hassnian, line 110: we should slowly get rid off the `rmrk` folder

components/massmint/types.ts
```diff
@@ -16,15 +16,15 @@ export type NFT = {
   description?: string
   price?: number
   status?: Status
-  attributes?: OpenSeaAttribute[]
+  tags?: OpenSeaAttribute[]
```
hassnian, line 19: why the change to `tags` I'd keep it as `attributes`

composables/massmint/parsers/parseTxt.ts
```diff
@@ -91,6 +94,7 @@ const updateEntries = (entries, block) => {
       file: entry.file,
       name: entry.name,
       description: entry.description,
+      tags: entry.attributes,
```
hassnian, line 97: `tags` and `attribute` let's keep one unless there is reason

hassnian (commented):
otherwise LGTM

components/massmint/EditPanel.vue
Jarsen136: done https://github.com/kodadot/nft-gallery/pull/11267/commits/71bbd1b82dcb50c4d443b7e537168faad8a8015e

composables/massmint/parsers/parseTxt.ts
Jarsen136: I guess the `attribute` is more user-friendly and easy to understand, so I put `attributes` file in the template file. However, when we call the polkdot API, we must use `tags` in the nft instead.

components/massmint/types.ts
Jarsen136: same reason as below comment

preschian (approved):

composables/massmint/parsers/parseTxt.ts
hassnian: Fair enough, but I would still keep one naming across the app **just for the sake of consistency** in this case attributes, since its already used in other places

See https://github.com/kodadot/nft-gallery/blob/main/components/common/EditNftModal.vue

And tags which is specific of SUB can be used inside transaction token mint logic

Let me know if it makes sense

Jarsen136: Alright, I have updated it.

hassnian (approved):

PR #11310, opened 2024-12-31
revert: downgrade devDeps #11289

preschian:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] deploy failed to netlify


preschian:
Confirmed, it's because of this PR: [Issue #11289](https://github.com/kodadot/nft-gallery/issues/11289). It's probably an error with Nuxt compatibility, specifically occurring on Netlify only

If we still want to use Netlify, we would need to revert it. I tried with Nuxt 3.15 as well, but it didn't help

cc @kodadot/internal-dev

Issue #11314, opened 2025-01-01
token `nextId` too large ?


hassnian:
### Is there an existing issue?

- [X] I have searched the existing issues

### What happened?

nextId seems to be the issue , @preschian can you have a look  

### How to reproduce

- try to create an offer for an item

![CleanShot 2025-01-01 at 09 53 03@2x](https://github.com/user-attachments/assets/e461aced-ad3a-4c5b-b408-33c4a8d6cd32)

### Please reproduce in steps

asd

### Expected Behavior

as

### Are you logged in?

None

### Screenshots

asd

### Payment link for reward

https://canary.kodadot.xyz/dot/transfer/?target=<My_Polkadot_Address_check_https://github.com/kodadot/nft-gallery/blob/main/REWARDS.md#creating-your-dot-address>

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

preschian:
checking

preschian:
when I close the modal, and try the offer button again looks fine. probably something conflicts on first modal poopup

hassnian:
> when I close the modal, and try the offer button again looks fine. probably something conflicts on first modal poopup

ok checking

hassnian:
@preschian try making this change, the nextId seems to be the issue 

![CleanShot 2025-01-01 at 10 23 53@2x](https://github.com/user-attachments/assets/090d4f47-7a90-4d4b-890b-784d1cfe501d)

preschian:
> @preschian try making this change, no error is thrown, the nextId seems to be the issue

noted

PR #11315, opened 2025-01-01
fix: ids too large

preschian:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context
- [x] Fix https://github.com/kodadot/nft-gallery/issues/11314. fix ids too large on assethub:


services/dyndata.ts
```diff
@@ -6,6 +10,42 @@ const api = $fetch.create({
...
+const lastIdAssethub = async (prefix: Prefix, collectionId: number) => {
+  const api = await apiInstanceByPrefix(prefix)
+  const ids = await api.query.nfts.item.keys(collectionId)
+  const ints = ids.map(id => Number.parseInt(id.toHuman()?.[1]?.replaceAll(',', '') || '0'))
+  const biggestId = Math.max(...ints)
+  return biggestId
+}
+
+// store latest IDs for each collection
+const latestIdsMap = new Map<string, number>()
```
hassnian, line 27:
```suggestion
```

will be slower but I'd remove this, with this implementation the map can become out of sync with the onchain data

services/dyndata.ts
```diff
@@ -6,6 +10,42 @@ const api = $fetch.create({
...
+
+export const generateIdAssethub = async (collectionId: number, prefix?: Prefix) => {
```
hassnian, line 29:
```suggestion
export const generateIdWith = async (collectionId: number, prefix?: Prefix) => {
```

hassnian, line 29: kinda verbose but something along those lines 

```suggestion
export const generateBackwardsCompatibilityId = async (collectionId: number, prefix?: Prefix) => {
```

services/dyndata.ts
```diff
@@ -6,6 +10,42 @@ const api = $fetch.create({
...
+  return ApiFactory.useApiInstance(endpoint)
+}
+
+const lastIdAssethub = async (prefix: Prefix, collectionId: number) => {
+  const api = await apiInstanceByPrefix(prefix)
+  const ids = await api.query.nfts.item.keys(collectionId)
+  const ints = ids.map(id => Number.parseInt(id.toHuman()?.[1]?.replaceAll(',', '') || '0'))
+  const biggestId = Math.max(...ints)
+  return biggestId
+}
```
hassnian, line 24: reuse/combine `getLastIndexUsedOnChain` in `composables/transaction/mintToken/utils.ts`

composables/drop/massmint/useDropMassMint.ts
```diff
@@ -18,7 +19,7 @@ export default () => {
   const tokenIds = ref<number[]>([])
   const populateTokenIds = async () => {
     for (const _ of Array.from({ length: amountToMint.value })) {
-      const tokenId = Number.parseInt(await generateId())
+      const tokenId = Number.parseInt(await generateIdAssethub(parseInt(drop.value.collection), usePrefix().urlPrefix.value as Prefix))
```
hassnian, line 22: is this not called for evm drops ?

services/dyndata.ts
```diff
@@ -6,6 +10,42 @@ const api = $fetch.create({
   baseURL: BASE_URL,
 })
 
+const apiInstanceByPrefix = (prefix: Prefix) => {
+  const endpoint: string = getChainEndpointByPrefix(prefix) || ''
+  return ApiFactory.useApiInstance(endpoint)
+}
```
hassnian, line 16: can we use `apiInstanceByPrefix` from `useApi`?

services/dyndata.ts
preschian: thank you, I use that one instead [3f58f61](https://github.com/kodadot/nft-gallery/pull/11315/commits/3f58f61bbe25bdefa0bcc8bbd8cdd957211ed0e2)

composables/drop/massmint/useDropMassMint.ts
preschian: should be under `if (isSub.value)`

services/dyndata.ts
preschian: this map for handling multiple ids, any alternative for that case?

hassnian: hmm i see , ideally some sort of subscription to keep that map synced would be needed

preschian: > hmm i see , ideally some sort of subscription to keep that map synced would be needed

Agree, but I think we need to go with that downside for this case.

If I summarize for the future improvements, currently there are several concerns with our genart for AssethHub:
- ID implementation during minting: [dyndata/src/index.ts#L15-L19](https://github.com/kodadot/private-workers/blob/de5b3cc38036621497eda233a51ad4dc99090e0f/services/dyndata/src/index.ts#L15-L19)
- Security flaws, which I previously mentioned here: [Issue #99](https://github.com/kodadot/private-workers/issues/99) (search "security issue" on that page)

To resolve both issues, we could use presigned mint, or something similar to lazy mint in EVM.

I haven't closely followed genart development since the beginning, so I'm not sure why presigned mint wasn't used. I believe there is a reason for that. Maybe others know the reason?

---

Anyway, I adjusted it a little bit: [`962b797`](https://github.com/kodadot/nft-gallery/pull/11315/commits/962b797a7a425546bd8ed4724a7521a6ac0afecc). For offer, it requires deployment from here first: [PR #234](https://github.com/kodadot/private-workers/pull/234)

preschian:
Hi team @kodadot/internal-dev, if there are no concerns, I recommend deploying this PR as a hotfix to production to support backward compatibility of IDs with older drops on AssethHub

I will write a postmortem for more details regarding this issue

hassnian (approved):

Jarsen136 (approved):

preschian:
merging this one

PR #11330, opened 2025-01-01
fix: stop user from inputting their own address

### LINKED ISSUE #11322: we should stop user from inputting their own address
image

autists like me will try it, its good that it does not show any nft in the list but the person should not get there at all

image

when inputted, put small text underneath the input as error: "You cannot enter your own address as the recipient"

Jarsen136:

## PR Type

- [x] Bugfix

## Needs Design check

- @exezbcz please review

## Needs QA check

## Context

- [x] Closes #11322

There is already an error handler for this case but occur another bug. Let me know if it's ok to keep the original error message and disabled button. @exezbcz

## Screenshot 📸

image

### DIFF
--- a/components/shared/AddressInput.vue
+++ b/components/shared/AddressInput.vue
@@ -7,7 +7,7 @@
     >
       <NeoInput
         v-model="inputValue"
-        :icon-right="iconRight"
+        :icon-right="inputIconRight"
         :placeholder="placeholder"
         data-testid="global-address-input"
         icon-right-clickable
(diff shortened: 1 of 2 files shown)


exezbcz:
please add also "x" icon to the input instead of the check that is there if everything is okay 


>There is already an error handler for this case but occur another bug. Let me know if it's ok to keep the original error message and disabled button. @exezbcz



ye its okay

Jarsen136:
> please add also "x" icon to the input instead of the check that is there if everything is okay
> 

updated

image

hassnian (approved):

vikiival (approved):

Issue #11268, opened 2024-12-19
Swaps wrong expiration time


hassnian:
![CleanShot 2024-12-19 at 08 47 48@2x](https://github.com/user-attachments/assets/37172338-37ba-43c9-aa2a-a1a3717dda49)

PR #11370, opened 2025-01-21
feat: expire trades based on `blockNumber`

hassnian:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

 atm there is a 1H delay between the time a trade is expired and when the indexer registers it 

so some users might be able to accept an already expired trade, this pr fixes that

CleanShot 2025-01-21 at 17 09 33@2x

- [x] #11268 (was not able to reproduce it but this should fix it )

## Screenshot 📸

CleanShot 2025-01-21 at 17 22 39@2x


Jarsen136 (approved):
Nice, work for me!

composables/useCurrentBlock.ts
vikiival, line 1: You can subscribe to blocks, 
probing rpc node each sec would imo result in 429

hassnian: > You can subscribe to blocks,
didn't know that, updated

hassnian:
> Otherwise ok; or [kodadot/stick#334](https://github.com/kodadot/stick/pull/334)

We still a 15min window that is not covered, so I'd have both

vikiival (approved):

Issue #9245, opened 2024-02-02
refactor: container size to tailwind


roiLeo:
Replace `container is-*` to tailwind equivalent

- ref #9013 
- https://bulma.io/documentation/layout/container/
- https://tailwindcss.com/docs/container

pbkompasz:
:wave:

kodabot:
ASSIGNED - @pbkompasz 🔒 LOCKED -> Sunday, February 4th 2024, 23:20:22 UTC -> 36 hours

pbkompasz:
@roiLeo Tailwind's container sets the `max-width` of the container to a given breakpoint e.g. anything width above 1536px, the `max-width` stays 1536px + the `.is-fluid` class adds 2.5rem padding on each side. 
Between breakpoints this happens:
![Screenshot from 2024-02-04 13-28-32](https://github.com/kodadot/nft-gallery/assets/47194071/af47e20a-2a67-4789-aa2c-96c266233a86)
![Screenshot from 2024-02-04 13-28-16](https://github.com/kodadot/nft-gallery/assets/47194071/1449a471-dc40-4ba1-9bf0-fe8a9d7e1508)
I could change it instead to `w-full px-[2.5rem]` and not use the container class.

roiLeo:
> @roiLeo Tailwind's container sets the `max-width` of the container to a given breakpoint e.g. anything width above 1536px, the `max-width` stays 1536px + the `.is-fluid` class adds 2.5rem padding on each side. 
> Between breakpoints this happens:
> ![Screenshot from 2024-02-04 13-28-32](https://github.com/kodadot/nft-gallery/assets/47194071/af47e20a-2a67-4789-aa2c-96c266233a86)
> ![Screenshot from 2024-02-04 13-28-16](https://github.com/kodadot/nft-gallery/assets/47194071/1449a471-dc40-4ba1-9bf0-fe8a9d7e1508)
> I could change it instead to `w-full px-[2.5rem]` and not use the container class.
> 

Let's try!

kodabot:
ASSIGNMENT EXPIRED - @pbkompasz has been unassigned.

zeshhaan:
Can share bit more context here as i look forward working on it again and make sure we are on the same page?

> Replace `container is-*` to tailwind equivalent
> 
> * ref [remove bulma css #9013](https://github.com/kodadot/nft-gallery/issues/9013)
> * https://bulma.io/documentation/layout/container/
[...]

Issue #9013, opened 2024-01-15
remove bulma css


preschian:
Related issue: https://github.com/kodadot/nft-gallery/issues/8071

PR #11391, opened 2025-02-04
🔧 .container.is-fluid to tailwind

roiLeo:
part of:
- #9013 
- #9245 

## Summary by CodeRabbit

- **Style**
  - Standardized container styling across pages and components by replacing inconsistent class names with `container-fluid`, resulting in a more uniform and full-width presentation.
  - Enhanced navigation bar layouts for a cleaner, more modern appearance with improved responsiveness on both desktop and mobile devices.
  - Introduced a new full-width styling utility, streamlining the overall design and ensuring a consistent user experience across the site.


hassnian (commented):
@roiLeo I think we should add `container-fluid`  as a utility class

https://v3.tailwindcss.com/docs/adding-custom-styles#adding-custom-utilities

i'ts much cleaner and more maintainable, wdyt?

components/Navbar.vue
```diff
@@ -8,11 +8,11 @@
     }"
   >
     <div
-      class="container items-center max-lg:!px-0"
-      :class="{ 'is-fluid': !isTouch }"
+      class="grow relative mx-auto my-0 max-lg:!px-0 items-stretch flex min-h-[3.25rem] w-full"
```
hassnian, line 11: Can't we reuse this ?

is the same as gallery-item-layout.vue

roiLeo: feel free to handle it

hassnian: what's the difference between this and container-fluid?

roiLeo:
> @roiLeo I think we should add `container-fluid` as a utility class
> 
> https://v3.tailwindcss.com/docs/adding-custom-styles#adding-custom-utilities
> 
> It's much cleaner and more maintainable, wdyt?

IMO, best solution would be to create a container wrapper component but that not the point of this PR.

preschian (approved):

hassnian (approved):

hassnian:
> > @roiLeo I think we should add `container-fluid` as a utility class
> > https://v3.tailwindcss.com/docs/adding-custom-styles#adding-custom-utilities
> > It's much cleaner and more maintainable, wdyt?
> 
> IMO, best solution would be to create a container wrapper component but that not the point of this PR.

i'm not opposed to that, feel free to raise a pr.

Jarsen136 (approved):

PR #11404, opened 2025-02-07
feat: fetch oda on drops page

preschian:
**Thank you for your contribution** to the [Koda - Generative Art Marketplace](https://kodadot.xyz).

👇 __ Let's make a quick check before the contribution.

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Refactoring

## Context

- [x] Requires deployment https://github.com/kodadot/private-workers/pull/241
- [x] Closes #11398
- [x] Closes #11400
- [x] Closes #11401

## Screenshot 📸

- [x] My fix has changed **something** on UI; a screenshot is best to understand changes for others.

| before | after |
|--------|--------|
| Screenshot 2025-02-07 at 14 22 19 | Screenshot 2025-02-07 at 14 22 21 |

PR #11405, opened 2025-02-10
fix: duplicate fetch on itemsgrid

preschian:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

## Context

- [x] followup: https://github.com/kodadot/nft-gallery/pull/11404
- [x] fix duplicate fetch on drops page
- [x] enable onchain data on collection page

### DIFF
--- a/components/items/ItemsGrid/useItemsGrid.ts
+++ b/components/items/ItemsGrid/useItemsGrid.ts
@@ -10,6 +10,7 @@ import type { NFTWithMetadata, TokenEntity } from '@/composables/useNft'
 import { nftToListingCartItem } from '@/components/common/shoppingCart/utils'
 import { useListingCartStore } from '@/stores/listingCart'
 import type { NFT, TokenId } from '@/types'
+import { fetchOdaToken } from '@/services/oda'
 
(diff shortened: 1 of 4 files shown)


components/items/ItemsGrid/useItemsGrid.ts
```diff
@@ -232,6 +235,46 @@ export function useFetchSearch({
     resetSearch()
   })
 
+  const processOnchainData = useDebounceFn(async () => {
+    items.value = await Promise.all(items.value.map(async (item) => {
+      if ('onchainData' in item && item.onchainData) {
```
Jarsen136, line 240:
```suggestion
      if (item.onchainData) {
```
It should be enough.

Jarsen136 (approved):
small stuff. otherwise lgtm

components/items/ItemsGrid/useItemsGrid.ts
preschian: need additional `'onchainData' in item` is because that custom key is doesn't exists in the types. I need to add to the types with `item.onchainData`. lmk which one do you prefer

PR #11418, opened 2025-02-18
fix: e2e test

Jarsen136:


Jarsen136:
@kodadot/internal-dev any idea maybe? On my localhost, these test pass normally.
image

PR #11578, opened 2025-05-30
feat(static): add `ahw` prefix

hassnian:
for the time being let's use ahw, since the new pallete revive is not available elsewhere 

@vikiival could you deploy


vikiival:
Will do

vikiival:
`+ @kodadot1/static@0.0.8-rc.0`

